PR #159, opened 2020-07-15
cloud-runner is fused with base image

DavidGOrtega:
Having too many images makes a bit confusing the message, thats specially true when you have to setup a self-hosted runner where you specify the image in the yaml and launch another image with docker.

To simplify it a bit I proposee to merge the CML cloud runner into the base image being
```dvcorg/cml``` equal to ```dvcorg/cml-cloud-runner```
```dvcorg/cml-gpu``` equal to ```dvcorg/cml-gpu-cloud-runner```
```dvcorg/cml-py3``` equal to ```dvcorg/cml-py3-cloud-runner```
```dvcorg/cml-gpu-py3``` equal to ```dvcorg/cml-gpu-py3-cloud-runner```

we still generate the images to as tags to not break workflows that may now be using the ```-cloud-runner``` tag. In the near future we can deprecate them totally.

@andronovhopf this PR implies also update the docs in [cml.dev](https://cml.dev) later on.

I wish we could do the same with py3 and gpu somehow.

### CODE REVIEW

**Comment by andronovhopf** [comment]
I like it a lot. See no problems on my end, once merged I'll update docs.

**Comment by andronovhopf** [comment]
What do you think about adding `py3` to the base image? Is there a strong reason not to?

**Comment by DavidGOrtega** [comment]
> What do you think about adding py3 to the base image? Is there a strong reason not to?

@andronovhopf py3 would be more difficult since people would tend to use the version they want. I mean

```sh
python ...
pip install ...
```

which version? I have always thought that the runner is unfortunately not well resolved from a GH  point of view. A GH user would maybe use the python action.

GPU is super doable we would just end up with a fatter image.

What do you think @andronovhopf and @dmpetrov ?

I think that this way it would less confusing, with just py3 we would have only:

 - cml
 - cml-py3

DavidGOrtega:
@dmpetrov only runner right? no gpu yet as far I understand

PR #179, opened 2020-07-27
Merge gpu into dockerfile

DavidGOrtega:
This PR is for sake of simplicity in CML message, managing a fatter image VS easier to be explained and less confusing.
We still generate ```cml-gpu``` and ```cml-gpu-py3``` for backward compatibility


elleobrien:
Looks awesome!

Issue #81, opened 2020-05-01
Python Action not supported


elleobrien:
I see in the README example how to install Python "manually", but there's a GH action `setup-python` already built for installing Python [that seems recommended](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions). 

I was trying this out with the cml docker container, and there seems to be an odd interaction between the docker container and `setup-python`.

For example, if I'm not using our docker container and run this workflow:
```yaml
name: your-workflow-name

on: [push]

jobs:
  run:
    runs-on: [ubuntu-latest]
    container: docker://dvcorg/dvc-cml:latest

    steps:
      - uses: actions/checkout@v2
      
      - name: Set up Python
        uses: actions/setup-python@v1
        with:
          python-version: '3.x'
          
      - name: Display Python version
        run: python -c "import sys; print(sys.version)"
      
      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip 
          pip install sklearn

```

Everything runs beautifully. As soon as I add `container: docker://dvcorg/dvc-cml:latest` to the .yaml file, though, I hit the following error:

![image](https://user-images.githubusercontent.com/29716724/80842455-77e49800-8bb6-11ea-8451-7ea804c32dfb.png)

Any ideas?

jbencook:
I got this working with the `dvcorg/cml:latest` image. By default, the container service mounts `/opt/hostedtoolcache` to `/__t` in the container, but that breaks the Python installs which are compiled to live in `/opt/hostedtoolcache` specifically.

But the container syntax lets you add mounted volumes. Here's a workflow file that worked for me:

```
name: test-setup-python

on: [push]

jobs:
  run:
    runs-on: [ubuntu-latest]
    container:
      image: docker://dvcorg/cml:latest
      volumes:
        - /opt/hostedtoolcache:/opt/hostedtoolcache

    steps:
      - uses: actions/checkout@v2
      
      - name: Set up Python
        uses: actions/setup-python@v1
        with:
          python-version: '3.x'
          
      - name: Display Python version
        run: python -c "import sys; print(sys.version)"
      
      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip 
          pip install sklearn
```

It also works with `setup-python@v2` and it works on a self-hosted runner. I'm pretty sure the base image will need to be Ubuntu 18.04, but it looks like all your images are so that wouldn't be a problem.

[...]

PR #185, opened 2020-07-29
Python setup action volume in workflow not needed

DavidGOrtega:
Thanks to @jbencook we were able to put back [Python setup](https://github.com/actions/setup-python) in shape.
However the volume had to be specified in the workflow #81 

This PR eliminates the need of having to setup the volume in the workflow ending with a simpler yaml:

```yaml
run:
    runs-on: [ubuntu-latest]
    container:
      image: docker://dvcorg/cml-dev:pyv10

    steps:
      - uses: actions/checkout@v2
      
      - name: Set up Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.x'
          
      - name: Display Python version
        run: python -c "import sys; print(sys.version)"
``` 

Works for GH runners and self-hosted runners.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -91,7 +91,7 @@ RUN apt update && \
 ENV RUNNER_PATH=/home/runner
 ENV RUNNER_ALLOW_RUNASROOT=1
 
-# FIXES Python Action
+# FIXES Python setup Action
 RUN mkdir /__t
 RUN ln -s /__t /opt/hostedtoolcache
 ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -90,7 +90,11 @@ RUN apt update && \
 # OUR DOCKER_MACHINE FORK SUPPORTS GPU ACCELARATOR
 ENV RUNNER_PATH=/home/runner
 ENV RUNNER_ALLOW_RUNASROOT=1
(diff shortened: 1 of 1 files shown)


DavidGOrtega:
Should we change the examples once its ready? @dmpetrov @elleobrien?

elleobrien:
I'm a little torn about which examples to change. Since this might break symmetry with GitLab (you'd still want Python in your Docker image for GitLab CI, since we don't have a comparable Python setup worked out for them yet).

So- should we change the GitHub code snippets only on cml.dev?  Likewise the sample project repos?
Or maybe we leave those as is, but release a blog and new sample project repo about using this improved method for Python setup that helps you do matrix testing? 

Thoughts?

dmpetrov:
Agree with @elleobrien - we should keep GH and GL parity.
It is great to have that working for users who prefer using python workflow. But let's keep using our dockers.

elleobrien:
OK, I'm going to merge now and we can think on how to create a way to showcase this. This is going to be relevant to a lot of users, knowing how thorny Python versions can get, so let's see if we can come up with a use case for a blog/video.

DavidGOrtega:
@elleobrien not breaking  symmetry with GitLab but being fully compliant with GH syntax is what I consider important. Users might use CML on existing workflows.
Additionally py3 is enforcing a python version while python setup does not.

I agree to not change the docs or tutorials but its a great addition to have this finally working! 😃

PR #194, opened 2020-08-03
use GITHUB_ACTIONS to check if running inside a Github Action

imrehg:
According to the docs at https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables the `GITHUB_ACTIONS` variable shows whether we are running inside and action. `GITHUB_ACTION` (without the plural) would contain the id of the action, which might not always be set (for example first time the action as added in a PR, and not on master). In the case when that variable is not set, commands try to use the gitlab setup erroneously. With this change the command will always find it being run in Github correctly.

Signed-off-by: Gergely Imreh <gergely.imreh@faculty.ai>

### DIFF
--- a/bin/cml-publish.js
+++ b/bin/cml-publish.js
@@ -6,35 +6,33 @@ console.log = console.error;
 const fs = require('fs').promises;
 const pipe_args = require('../src/pipe-args');
 const yargs = require('yargs');
-const {publish_file} = require('../src/report');
+const { publish_file } = require('../src/report');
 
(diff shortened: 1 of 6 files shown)


imrehg:
Sorry, looks like eslint and restyled is playing ping-pong with the styling, or maybe I'm doing something silly? How is the best to go about this?

DavidGOrtega:
@imrehg thanks for the contribution!
Tests are not passing due to credentials, we have an open ticket for that.
Lets merge in ```contrib``` to be able to run tests and then I merge master

PR #197, opened 2020-08-03
use GITHUB_ACTIONS to check if running inside a Github Action (#194)

DavidGOrtega:
via #194 

According to the docs at https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables
the `GITHUB_ACTIONS` variable shows whether we are running inside
and action. `GITHUB_ACTION` (without the plural) would contain the
id of the action, which might not always be set (for example
first time the action as added in a PR, and not on master). In
the case when that variable is not set, the commands try to use the
gitlab setup erroneously. With this change the command will
always find it being run in Github correctly.

Signed-off-by: Gergely Imreh <gergely.imreh@faculty.ai>

### DIFF
--- a/bin/cml-publish.js
+++ b/bin/cml-publish.js
@@ -8,7 +8,7 @@ const pipe_args = require('../src/pipe-args');
 const yargs = require('yargs');
 const { publish_file } = require('../src/report');
 
-const { handle_error } = process.env.GITHUB_ACTION
+const { handle_error } = process.env.GITHUB_ACTIONS
   ? require('../src/github')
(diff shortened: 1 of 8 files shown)


DavidGOrtega:
@imrehg we continue here

imrehg:
Built the CML Docker image and tested it on a new repo (where I hit original issue as well), and it seems work as expected.

DavidGOrtega:
Thanks for the contribution @imrehg ! 🙇

PR #201, opened 2020-08-06
Support for SVG mime type

### LINKED ISSUE #200: cml-publish: fails on SVG files
Due to this problem https://github.com/sindresorhus/file-type/issues/58 we can't detect SVG mime and the pipeline fails:

```
  dvc plots show --show-vega prc.json > vega.json
  vl2svg vega.json prc.svg
  
  echo "## Metrics report" > report.md
  cml-publish prc.svg --md >> report.md
  cml-send-comment report.md
  shell: sh -e {0}
  env:
    repo_token: ***
npx: installed 58 in 3.731s
```

with a cryptic error:

```
Cannot destructure property 'mime' of '(intermediate value)' as it is undefined.
```

- we should handle errors better
- SVG is quite common and useful since it's the easiest way to solve #99

DavidGOrtega:
Adds support for SVG files in the ```cml-publish``` command. Adds also a better error handler when mime type guess fails.

closes #200


src/utils.js
```diff
@@ -19,24 +20,45 @@ const exec = async (command, opts) => {
   });
 };
 
+const mime_type = async opts => {
+  const { path, buffer } = opts;
+
+  try {
+    const svg_cadidate = path
```
shcheklein, line 27: typo: cadidate -> candidate

shcheklein: I wonder if we should just rely on a file extension instead of reading it? (if it's a path) ... just to avoid reading into memory. And potentially it should be enough, I guess.

DavidGOrtega: > wonder if we should just rely on a file extension instead of reading it?

I have added that check to not have to read it in memory

PR #206, opened 2020-08-06
Cml action

DavidGOrtega:
This PR introduces CML as a GitHub Action.
As discussed with @dmpetrov introducing a single action for a multi command package enforces us to build a monstrous GitHub Action with dozens of parameters. 
Following the example of Hashicorp's Terraform with their actions: 
 - [Terraform GitHub Actions](https://github.com/hashicorp/terraform-github-actions)
 - [Setup Terraform](https://www.terraform.io/docs/github-actions/setup-terraform.html)

we can conclude that our assumptions are right and the best way to  do this is to generate an action that will install CML along with DVC (this does not means that DVC should have its own action independently).

```yaml
jobs:
  check:
    runs-on: [ubuntu-latest]
    
    steps:
    - uses: actions/checkout@v2

    - uses: iterative/cml
      with:
        version: latest
        dvc_install: true
        dvc_version: latest
```


dmpetrov:
> That's amazing. Could you please make a separate action for DVC? I created a repo for that: https://github.com/iterative/dvc-action

Also, we might need separate actions for clouds (might be for each cloud separately). Let's make it in another iteration/PR.

PR #223, opened 2020-08-20
Docker layer cache

DavidGOrtega:
This PR introduces docker layer cache. The main goal is to accelerate the use of CML images making Github caching the layers, however looks like the cache is only effective from the repository. Worst case scenario is at least accelerate the ```check-container``` workflow test and the process of building CML images.


DavidGOrtega:
Performance of GH cache is awful. It adds 9 extra minutes to the process

PR #236, opened 2020-09-04
Fix cml-comment pull request sha

### LINKED ISSUE #222: cml-send-comment pull request events is not effective on GH
Pull request UI is not showing commits that happens on PR commits; It's only showing the commits that happens in the branch. Because of that ```cml-send-comment ``` does not apparently work. It does work but is commenting on a commit that is not displayed in the UI

DavidGOrtega:
Pull requests in GH is pointing the sha to the latest commit of the PR branch not even displaying the comment that the user would make in that commit.

closes #222


DavidGOrtega:
It works now, having two comments one for push and one for pull_request, the "strange" thing is that the UI is merging both as seen below, while the there are two independent comments that can be seen [here](https://github.com/iterative/cml/commit/f47691f4ae6ef3cec4c38c7e019fd8173fa80480)

bin/cml-send-comment.test.js
```diff
@@ -30,9 +30,9 @@ describe('CML e2e', () => {
       "Usage: cml-send-comment.js <path> --head-sha <string>
 
       Options:
-        --version   Show version number                                      [boolean]
-        --head-sha  Commit sha
-        -h          Show help                                                [boolean]"
+        --version     Show version number                                    [boolean]
+        --commit-sha  Commit sha
```
dmpetrov, line 34: This looks like an API change. It is better to keep the old option for backward compatibility with some warning messages and depreciate later (in a few months).

dmpetrov (approved):
👍 One comment re the compatibility.

bin/cml-send-comment.test.js
DavidGOrtega: sounds good

PR #240, opened 2020-09-04
Removal of vega

DavidGOrtega:
We still had vega in the stack because of a function. 
This PR removes vega from CML but not from the CML image.

Additionally puts back ```prettier``` in dev dependencies.


dmpetrov (approved):
nice cleanup

PR #270, opened 2020-09-25
Cml cloud runner param gpu

### LINKED ISSUE #260: cml-cloud-runner detect gpu using nvidia-smi
Before starting the runner we can check  if ```-gpus all``` is allowed using ```nvidia-smi```

DavidGOrtega:
Removes gpu param using nvidia-smi. Also checks if docker runner was able to succesfully start.

closes #260


dmpetrov (approved):
Looks good

PR #277, opened 2020-09-29
Add support for Github organisation level runners

JamesMowatt:
This PR adds support for organisation level runner registration in Github, you are able to register a runner that can be used by multiple repositories.

### DIFF
--- a/src/github.js
+++ b/src/github.js
@@ -74,15 +74,13 @@ const get_runner_token = async () => {
     return token;
   }
 
-  if (typeof repo === 'undefined') {
-    const {
-      data: { token }
-    } = await octokit.actions.createRegistrationTokenForOrg({
-      org
-    });
+  const {
+    data: { token }
+  } = await octokit.actions.createRegistrationTokenForOrg({
+    org
+  });
 
-    return token;
-  }
+  return token;
 };
 
 const register_runner = async (opts) => {
--- a/src/github.js
+++ b/src/github.js
@@ -11,6 +11,7 @@ const {
 } = process.env;
 
 const [owner, repo] = GITHUB_REPOSITORY.split('/');
+const org = owner;
 const IS_PR = GITHUB_EVENT_NAME === 'pull_request';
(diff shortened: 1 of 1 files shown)


src/github.js
```diff
@@ -62,14 +63,26 @@ const comment = async (opts) => {
...
 const get_runner_token = async () => {
-  const {
-    data: { token }
-  } = await octokit.actions.createRegistrationTokenForRepo({
-    owner,
-    repo
-  });
-
-  return token;
+  if (typeof owner !== 'undefined' && typeof repo !== 'undefined') {
```
DavidGOrtega, line 66: Lets remove this if putting the block below the organization code 

```js
if (typeof owner !== 'undefined' && typeof repo === 'undefined') {
//do
}


```

DavidGOrtega, line 66: typeof owner !== 'undefined' does not have to be checked since its a condition of both

DavidGOrtega (changes requested):
Lets remove this if moving this block after the organization check if block

```js
if (/*is organization */) {
//do
}




```

DavidGOrtega:
@JamesMowatt The idea is to pass  ```--RUNNER_REPO="https://github.com/myorg"``` ?

DavidGOrtega:
Looks great! Just only minor issues @JamesMowatt 

Something missing is Gitlab. Same functionality has to be offered. 
Would be possible for you to do it? If not no worries.

JamesMowatt:
> @JamesMowatt, thanks for the help! 
> 
> The idea is to pass `--RUNNER_REPO="https://github.com/myorg"` ?

Yes, you would only pass the organisation.



> Looks great! Just only minor issues @JamesMowatt
> 
> Something missing is Gitlab. Same functionality has to be offered.
> Would be possible for you to do it? If not no worries.

I will look to raise the Gitlab work as a separate PR if that is ok?

> Lets remove this if moving this block after the organization check if block
> 
> ```js
> if (/*is organization */) {
> //do
> }
> ```

I have applied these requested changes.

DavidGOrtega:
Looks perfect!

PR #296, opened 2020-10-08
Bitbucket integration

elleobrien:
To do:

- [ ] Logic for detecting CI platform (GitLab, GH, or BB) in `bin/cml-send-comment.js` needs to be decided
- [x] Check tensorboard/cml-publish functions
- [ ] Need a code quality and style check by David!


bin/cml-send-comment.js
```diff
@@ -7,7 +7,7 @@ const yargs = require('yargs');
 
 const { head_sha: HEAD_SHA, handle_error, comment } = process.env.GITHUB_ACTIONS
   ? require('../src/github')
-  : require('../src/gitlab');
+  : require('../src/bitbucket');
```
DavidGOrtega, line 10: we have to find a way to determine the env. However this might be useless with the new approach

package.json
```diff
@@ -1,10 +1,7 @@
 {
   "name": "@dvcorg/cml",
   "version": "0.1.49",
-  "author": {
-    "name": "DVC",
-    "url": "http://cml.dev"
-  },
+  "author": "DVC (http://cml.dev)",
```
DavidGOrtega, line 4: just curiosity, why do you like this format?

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
+const fetch = require('node-fetch');
+const FormData = require('form-data');
```
DavidGOrtega, line 2: unused

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+const { fetch_upload_data } = require('./utils');
```
DavidGOrtega, line 3: unused

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+const {Base64} = require('js-base64');
+
+const {
+  BITBUCKET_REPO_FULL_NAME = '', //namespace with project name, i.e. elle/cml
+  BITBUCKET_REPO_UUID, // UUID of project- string in {}
```
DavidGOrtega, line 8: unused

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+  BITBUCKET_BRANCH, // branch, cannot do build against tag
+  BITBUCKET_TOKEN, // token, cannot do build against branch
+  BITBUCKET_COMMIT, // SHA of commit
+  BITBUCKET_PR_ID, // ID of pull request- only available if build is triggered on PR
+  // GITLAB_USER_EMAIL, // doesn't look like email is part of BB env vars
+  BITBUCKET_WORKSPACE, //username
+  BITBUCKET_GIT_HTTP_ORIGIN, // url to project
```
DavidGOrtega, line 15: unused

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+const USERNAME = BITBUCKET_WORKSPACE;
+const IS_PR = BITBUCKET_PR_ID;
+const REF = BITBUCKET_BRANCH || BITBUCKET_TOKEN;
+const PASSWORD = repo_token;
+const API_URL = "https://api.bitbucket.org/2.0";
+
+const comment = async (opts) => {
+  const { commit_sha, report } = opts;
+
+  const endpoint = `/repositories/${USERNAME}/${BITBUCKET_REPO_SLUG}/commit/${BITBUCKET_COMMIT}/comments/`;
```
DavidGOrtega, line 30: should be using commit_sha, the cml command will set it with BITBUCKET_COMMIT by default since you are exposing BITBUCKET_COMMIT later on as head_sha ;)

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+exports.ref = REF;
+exports.head_sha = BITBUCKET_COMMIT;
+exports.user_name = USERNAME;
+exports.comment = comment;
+exports.get_runner_token = get_runner_token;
+exports.handle_error = handle_error;
+exports.token = PASSWORD;
+exports.repo = BITBUCKET_REPO_SLUG;
+
+exports.CHECK_TITLE = 'CML Report';
```
DavidGOrtega, line 88: needs extra line

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+
+  if (!endpoint) throw new Error('BitBucket API endpoint not found');
+
+  const headers = {
+            'Content-Type': 'application/json',
+            'Authorization': 'Basic ' + Base64.encode(`${USERNAME}:${PASSWORD}`)
+        };
+                    
+  
+  console.log(`${API_URL}${endpoint}`)
```
DavidGOrtega, line 58: console is intended?

src/bitbucket.js
```diff
@@ -0,0 +1,88 @@
...
+  
+  const response = await fetch(`${API_URL}${endpoint}`, {
+        method,
+        headers,
+        body
+  });
+  const json = await response.json();
+
+  console.log(json);
```
DavidGOrtega, line 67: console

DavidGOrtega (commented):
Looks very good!

Just some comments. Looks like you have issues with eslint, looks like is not working pre-commiting the work. 
There is a tool called husky that is running before the commit and fixing and prettifying the code. Also exposing errors.
Let me show it to you later if you need.

image

DavidGOrtega (changes requested):
Until we fix the import we cant merge this

src/drivers/bitbucket_cloud.js
```diff
@@ -0,0 +1,79 @@
...
+  async runner_by_name(opts = {}) {
+    throw new Error('BitBucket does not support runner_by_name!');
+  }
+
+  async request(opts = {}) {
+    const { token, api } = this;
+    const { endpoint, method = 'GET', body } = opts;
+
+    if (!endpoint) throw new Error('BitBucket API endpoint not found');
+    const {BITBUCKET_WORKSPACE}=process.env;
```
DavidGOrtega, line 66: This makes bitbucket dependant on Env again. If we are in gitlab CI pushing to Bitbucket would need to setup this ENV variable previously.

DavidGOrtega, line 66: This makes bitbucket dependant on Env again. If we are in gitlab CI pushing to Bitbucket would need to setup this ENV variable previously.

src/drivers/bitbucket_cloud.js
```diff
@@ -0,0 +1,79 @@
...
+    const headers = {
+      'Content-Type': 'application/json',
+      'Authorization': 'Basic ' + Base64.encode(`${BITBUCKET_WORKSPACE}:${token}`)};
```
DavidGOrtega, line 69: Lets require the token full. I mean the user has to provide is the ```repo_token``` or ```BITBUCKET_TOKEN``` as env or as a parameter and thats the base64 of their user and password. They have to calculate manually but I think is fine. What do you think?

```js
const headers = {
      'Content-Type': 'application/json',
      'Authorization': `Basic ${token}`
};
```

I have thought also to extend the input with ```user``` ```pass``` alternatively to ```token``` so we can do it for the user but Im not sure yet. What do you think?

package.json
```diff
@@ -57,11 +57,13 @@
     ]
   },
   "dependencies": {
+    "@actions/core": "^1.2.5",
```
DavidGOrtega, line 60: This is not needed

DavidGOrtega (changes requested):
Amazing @elleobrien!

Just only one consideration about the token and ENV dependancy.

src/drivers/bitbucket_cloud.js
elleobrien: Hm, if they're in GitLab CI trying to make a request to the BBCloud api, we need some way of getting their BB username. 

We could try using the repo URL, but if they're using an organizational workspace the URL won't necessarily contain their account name. What makes sense?

Assuming we let folks add their username as a secret (when not using BBCloud as driver), I imagine we can do:
```
if (type(repo) == BBCloud){
 const {BITBUCKET_WORKSPACE}=process.env;
 const username = BITBUCKET_WORKSPACE
} else{
 const {BITBUCKET_ACCOUNT_SECRET} = process.env;
 const username = BITBUCKET_ACCOUNT_SECRET
};

```

src/drivers/bitbucket_cloud.js
elleobrien: so you are saying that the user will need to manually generate a token that is 

base64($username:$password)? 

The only concern i have is... will this force them to write a little base64 encoder, or use some online widget to do it? I don't like asking people to manually create the key. I imagine a lot of users getting lost at this step or quitting, unless they'res an easy way to do I don't know. 

Not sure I understand your idea about extending the input, maybe if you make some sample code I could understand better!

package.json
elleobrien: thanks for catching

src/drivers/bitbucket_cloud.js
DavidGOrtega: @elleobrien 
instead of using token they would have to setup user and password like this

```sh
cml-send-comment --user $MYUSER --pass $MYPASS report.md
```

alternatively they could set 
```sh
cml-send-comment --token $MYTOKEN report.md
```

where token is the base64($username:$password)

src/drivers/bitbucket_cloud.js
DavidGOrtega: > we need some way of getting their BB username.

As input or pre calculated token see below.
We can't depend on variables. The new architecture tries to infer as much as it can but now this driver wont operate in Gitlab CI without setting and ENV var

src/drivers/bitbucket_cloud.js
dmpetrov: I'm trying to understand what is the fundamental issue here. Why BB cannot be implemented the same way as GH/GL with a single token? Will the problem be solved by the separation of BB-cloud & BB-server drivers? I'd appreciate it if you can explain it.

Passwords and secrets in the command line is a potential security breach. In some settings, these passwords will end up in a centralized logging system with wide access across org. It needs to be avoided as much as possible.

dmpetrov (changes requested):
It is an absolutely amazing change ✨ CML is "escaping" GH/GL.

I have a question about token(s) and passwords - see inline

elleobrien:
To answer your question @dmpetrov, BB doesn't provide "personal access
tokens". You need to authenticate with username + password. BUT you can add
those as "masked" secrets, so they're not saved in any logging (i.e., the
BitBucket pipeline logs). So it's no more a security breach than any other
secret.

I think I don't understand what is the problem with using an if-statement.
If the driver is inferred to be BitBucket cloud, we can use an
environmental variable. If not, we can look for user credentials as secrets
(username & password, or one token- base64(username:password). What would
the issue be with that approach?

DavidGOrtega:
@dmpetrov as @elleobrien the user pass or token has to be secrets.
The main issue right now with this PR is that the request is generating the token from an ENV VAR and we dont want to do that anymore. Doing that implies that we have to setup that VAR in another CI ENV or local.

```js
Authorization': 'Basic ' + Base64.encode(`${BITBUCKET_WORKSPACE}:${token}`)};
```

So the solution is expect the token to be generated by them as base64 of user:pass like this:
```
$(echo  $USER_SECRET:$PASS_SECRET | base64)
```

Or introduce the possibility os setting ```user``` and ```password```.
of course @elleobrien we could infer the ```user``` from ```BITBUCKET_WORKSPACE``` but now we are just only reading it from ENV always. If I were using GL as CI I can not comment without setting BITBUCKET_WORKSPACE ENV  VAR

elleobrien:
> @elleobrien of course we could infer the `user` from `BITBUCKET_WORKSPACE` but now we are just only reading it from ENV always. If I were using GL as CI I can not comment without setting BITBUCKET_WORKSPACE ENV VAR.

@DavidGOrtega yes I got this- what I'm wondering is, should we use an if-statement to gate this? So...

if inferred driver != BB:
   look for user ID secret
if inferred driver == BB:
  use environmental variable

The idea being, don't make people add a secret if they don't need to (when BB is the driver). It does look like the GH & GL modules use some environmental variables, but based on what the inferred driver is. 

It sounds like the alternative is that we always make people specify the username & password when they want to leave CML comments on a BB PR, regardless of their driver. Is this what you think is right, @DavidGOrtega? Apologies if I'm not understanding!

DavidGOrtega:
@elleobrien yeah I think that the saver way to do this is just enforce them to setup a token like always with an example.
 
they can generate it with

```
echo  $USER_SECRET:$PASS_SECRET | base64
```

so they just setup token or pass token this way

src/cml.js
```diff
@@ -35,12 +39,13 @@ const get_driver = (opts) => {
...
   if (driver === 'github') return new Github({ repo, token });
   if (driver === 'gitlab') return new Gitlab({ repo, token });
+  if (driver === 'bitbucket') return new BitBucketCloud({ repo, token });
 
   throw new Error(`driver ${driver} unknown!`);
 };
 
 const infer_token = () => {
-  const { repo_token, GITHUB_TOKEN, GITLAB_TOKEN } = process.env;
+  const { repo_token, GITHUB_TOKEN, GITLAB_TOKEN} = process.env;
```
DavidGOrtega, line 48: lets add BITBUCKET_TOKEN to maintain the structure

src/drivers/bitbucket_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+    const { repo, token } = opts;
+
+    if (!token) throw new Error('token not found');
+    if (!repo) throw new Error('repo not found');
+
+    this.token = token;
+    this.repo = repo;
+
+    const { protocol, host, pathname } = new URL(this.repo);
+    this.repo_origin = `${protocol}//${host}`;
```
DavidGOrtega, line 18: double slash

src/drivers/bitbucket_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+    const output = await this.request({ endpoint, method: 'POST', body });
+
+    return output;
+  }
+
+  async check_create() {
+    throw new Error('BitBucket Cloud does not support check!');
+  }
+
+  async upload(opts = {}) {
```
DavidGOrtega, line 40: Does BB support any kind of upload?

DavidGOrtega (changes requested):
Perfect! Just only three silly things

src/drivers/bitbucket_cloud.js
elleobrien: there are artifacts, but they're impermanent- 14 days https://support.atlassian.com/bitbucket-cloud/docs/use-artifacts-in-steps/
Makes me think there are very few advantages to integrating with CML.

src/drivers/bitbucket_cloud.js
elleobrien: is double slash wrong? I assumed it's an escape

elleobrien: you're right, it's fine to use single slash. but note that i had copied this code from `gitlab.js`- so same double slash is there

src/cml.js
elleobrien: done!

src/drivers/bitbucket_cloud.js
DavidGOrtega: Ah sorry its OK!!!! protocol needs double slash!

DavidGOrtega (approved):

DavidGOrtega:
Excellent! 🚀

PR #301, opened 2020-10-09
Watermark for viewer

### LINKED ISSUE #285: CML watermark for the comment
We need that to deduplicate the CML comments and import the CML reports from external tools

DavidGOrtega:
This PR introduces a watermark in the comment to be able for the viewer to distinguish between a CML report and the rest.

closes #285


DavidGOrtega:
I think that the idea works pretty nice, however I still think that the svg is way better, maybe if we put it in the repo and referencing it as raw? That would work and make also people trust the image.

Current approach downside is that in bitbucket does not work.

DavidGOrtega:
```
![CML watermark](https://raw.githubusercontent.com/iterative/cml/watermark-comment/assets/watermark.svg)
```
What do you think? It works on bitbucket too!!!

CML watermark

dmpetrov (approved):
LGTM

PR #311, opened 2020-10-14
CML contexts

DavidGOrtega:
This PR introduces contexts, this is the ability to be able to setup a repo and token to the command, so you can run under gitlab CI and comment in GH scm setting the repo and token in the command.
Additionally the repo and sha is taken from the repo itself making future integrations faster if the CI do a full clone of the repo instead of a mirror like GL does with gitlab.

Benefits: 
 - Integrations are easier since we are not looking for specific CI ENV variables but extracting the repo, sha, etc... from the repo.
 - Commands are independent to be run locally or in a non CI environment.
 - We could add another kind of drivers, slack, confluence, etc... 
 - Very easy to work with you on premise instances.

### Example of a GL runner commenting under a GH repo setting repo_token to GH. 

Note that token and commit-sha is not needed since the first can be extracted from the env variables (secret) and the second is inferred from the repo itself.:

```yaml
train:
  script:
    - echo "Training..."
    - echo 'Hi from cml!' >> report.md
    - cml-send-comment --repo=https://github.com/iterative/3_tensorboard  report.md
    - echo "Trained!"
```

### example of your own github on premise:
```yaml
train:
  script:
    - echo "Training..."
    - echo 'Hi from cml!' >> report.md
    - cml-send-comment --repo=https://myentreprisegithub.com/iterative/3_tensorboard  --driver=github report.md
    - echo "Trained!"
```


bin/cml-cloud-runner-entrypoint.js
```diff
@@ -1,40 +1,40 @@
...
-  RUNNER_REPO,
   RUNNER_IDLE_TIMEOUT = 5 * 60,
   RUNNER_LABELS = 'cml',
   RUNNER_NAME = randid(),
   RUNNER_EXECUTOR = 'shell',
   RUNNER_RUNTIME = '',
-  RUNNER_IMAGE = 'dvcorg/cml:latest'
+  RUNNER_IMAGE = 'dvcorg/cml:latest',
+
+  RUNNER_DRIVER,
```
elleobrien, line 19: It seems like the flags and some js modules use `driver`, but others use `client`- if I understand correctly, they're referring to the same thing. Is this intentional?

src/cml.js
```diff
@@ -0,0 +1,95 @@
...
+    const client = get_client(this);
+    return await client.check_create(opts);
+  }
+
+  async publish(opts = {}) {
+    const { title = '', md, gitlab_uploads } = opts;
+
+    let mime, uri;
+
+    if (gitlab_uploads) {
```
elleobrien, line 65: Is there any way to move the gitlab_uploads logic out of `CML.publish()`? The only reason I ask is that if we expand to more clients with their own unique methods for uploading artifacts, I could imagine this logic getting hard to maintain.

elleobrien (commented):
@DavidGOrtega looks pretty nice- everything makes sense. Two big comments to discuss:

1. I don't think I fully understand where we would fit in the code for, say, Jenkins as my CI server- right now, the client code looks for environmental variables that are special to their runners. But I'm trying to imagine what the logic would be for Bitbucket with a Jenkins CI integration- since it looks like registering runners, creating comments, and calling the API all still happens in the same module as runner environmental variables are used (`GitlabClient.js` and `GithubClient.js`). How should we think about that kind of development?

2. Because we define properties and methods of the clients in `GitlabClient.js` and `GithubClient.js`, and then _again_ define those same methods and clients in `cml.js` (although in a very high-level way), it almost seems like there's duplicated code. It makes very clear sense why this pattern is used, but I'm concerned about maintenance with two extremely similar "levels". Do you think this is a concern?

Nice work, and looking forward to chatting at our meeting!

.github/workflows/publish.yml
```diff
@@ -4,6 +4,12 @@ on: [push, pull_request]
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+  TEST_GITHUB_TOKEN: ${{ secrets.TEST_GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
DavidGOrtega: Yes, in order to test it independently we need to setup env variables
```
  TEST_GITHUB_TOKEN: ${{ secrets.TEST_GITHUB_TOKEN }} 
  TEST_GITHUB_REPO: https://github.com/iterative/cml_qa_tests_dummy/
  TEST_GITHUB_SHA: 62edc8b3f46a60b3fe1e5c08fd3e0046d350ee29
  TEST_GITLAB_TOKEN: ${{ secrets.TEST_GITLAB_TOKEN }}
  TEST_GITLAB_REPO: https://gitlab.com/iterative.ai/cml_qa_tests_dummy/
  TEST_GITLAB_SHA: c4c13286e78dc252dd2611f31a755f10d343fbd4
```

bin/cml-cloud-runner-entrypoint.js
```diff
@@ -1,40 +1,40 @@
...
+  repo_token
 } = process.env;
 
-const { protocol, host, pathname } = new URL(RUNNER_REPO);
+const cml = new CML({
+  driver: RUNNER_DRIVER,
+  repo: RUNNER_REPO,
+  token: repo_token
+});
+const IS_GITHUB = cml.driver === 'github';
```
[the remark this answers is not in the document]
DavidGOrtega: lines 68, 115 and 150 in the same file

src/cml.js
```diff
@@ -0,0 +1,95 @@
...
+  throw new Error('driver unknown!');
+};
+
+class CML {
+  constructor(opts = {}) {
+    const env_driver = () => {
+      const { repo } = opts;
+      const { GITHUB_REPOSITORY, CI_PROJECT_URL } = process.env;
+
+      if (repo && repo.startsWith('https://github.com')) return 'github';
```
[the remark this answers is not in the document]
DavidGOrtega: Yes, we can actually infer the driver automatically in the environment or even setup the driver, repo and token to be pointing  whatever  we want

src/cml.js
```diff
@@ -0,0 +1,95 @@
...
+      if (repo && repo.startsWith('https://gitlab.com')) return 'gitlab';
+
+      if (GITHUB_REPOSITORY) return 'github';
+      if (CI_PROJECT_URL) return 'gitlab';
+    };
+
+    const { driver = env_driver(), repo, token } = opts;
+    this.driver = driver;
```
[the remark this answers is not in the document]
DavidGOrtega: Points to the object itself,  its the  classic OOP you know

bin/cml-cloud-runner-entrypoint.js
DavidGOrtega: I dont remember to have set any client as a parameter, however you are right the ```driver``` points to one of the clients GitlabClient or GitlabClient.
Im changing them  for Gitlab and GitHub and are going to be inside the drivers folder. To be more clear

src/cml.js
DavidGOrtega: I think this is where it should go actually, the real logic is in the client already which is calling the gitlab api but here we define the business logic somehow you have to decide which upload is going to be used cml's native one or one provide by the client/driver

src/cml.js
```diff
@@ -0,0 +1,95 @@
...
+  const { driver, repo, token } = opts;
+  if (!driver) throw new Error('driver not set');
+
+  if (driver === 'github') return new GithubClient({ repo, token });
+  if (driver === 'gitlab') return new GitlabClient({ repo, token });
+
+  throw new Error('driver unknown!');
+};
+
+class CML {
```
dmpetrov, line 15: Is CML (and `cml.js`) the right name? The code ends up using CML for different objects like:
```
const CML = require('../src/cml');
…
const cml = new CML(opts);
```

One of the big ideas behind this redesign is to separate different concepts - GitHub/GitLab and CI/CDs executors. But it seems like we still using a monolithic CML concept.

I'd suggest renaming this one to something like `ScmHostingDriver` with the factory method and abstract (empty) methods (or without methods since it is JS).

src/cml.js
```diff
@@ -0,0 +1,95 @@
...
+  async publish(opts = {}) {
+    const { title = '', md, gitlab_uploads } = opts;
+
+    let mime, uri;
+
+    if (gitlab_uploads) {
+      const client = get_client({ ...this, driver: 'gitlab' });
+      ({ mime, uri } = await client.publish(opts));
+    } else {
+      ({ mime, uri } = await upload(opts));
```
dmpetrov, line 69: It seems like there is one case is not covered - `publish(gitlab_uploads=false)` for GitLab client.

src/cml.js
dmpetrov: If this CML is a general, user-facing API - it makes sense to keep it here and keep the name. But all the client's specific logic needs to be extracted. Including the abstract `*Driver` class with the factory method.

dmpetrov (changes requested):
The change makes the existing API cleaner. But I still see some opportunities in making API even better.

In the change. it tries to abstract out the providers (GH/GL) from logic but keep the separation on the public API level (CML class). It looks like we need to separate the public class (CML), the providers/drivers logic (it is mostly done), and the installation logic (this is part of CML for some reason).

Now: CML API (with some logic) --> Providers (with logic)
Should be: CML API (no logic, dummy proxy)--> Providers Factory (no logic, just abstracting out the providers) --> Providers(all logic is here)

Also, it is not clear from the change on how to use different systems:
1. how to work with GitHub provider but GitLab CI/CD?
2. how we suppose to integrate Jenkins (CI/CD) with GitLab/GitHub provider?

src/cml.js
DavidGOrtega: @dmpetrov it is. The logic to publish in Gitlab is inside the Gitlab client. In CML there is only a decision. Take into account that even of you work in Gitlab you can still publish with CML storage. So its just an if

```js
if (gitlab_uploads) {
      const client = get_client({ ...this, driver: 'gitlab' });
      ({ mime, uri } = await client.publish(opts));
    } else {
      ({ mime, uri } = await upload(opts));
    }
```

If you see all the specific publish code is inside the Gitlab client

DavidGOrtega: If we were using that possibility we would be adding the  the CML upload in every client

```js
} else {
      ({ mime, uri } = await upload(opts));
    }
```

src/cml.js
DavidGOrtega: publish(gitlab_uploads=false)  goes to default upload the else

src/cml.js
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    const { driver = env_driver(), repo, token } = opts;
+    this.driver = driver;
+    this.repo = repo;
+    this.token = token;
+  }
+
+  env_repo() {
+    return get_client(this).env_repo();
```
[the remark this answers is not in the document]
DavidGOrtega: Its not a factory but a facade. CML api its defining some methods and then its calling the proper underlying code. That code is wrapped inside every client.

I dont see code duplication but expose the methods that it should expose. Users that would use the CML SDK should not dig into the drivers/clients. Just only know that several operations are available in the CML.

DavidGOrtega:
@dmpetrov 

> Now: CML API (with some logic) --> Providers (with logic)
Should be: CML API (no logic, dummy proxy)--> Providers Factory (no logic, just abstracting out the providers) --> Providers(all logic is here)

Its a facade. All the business logic (input data check, output data, etc...) should be done in the Facade Class, also its the one that exposes and defines all the methods available. It could be the class to be exposed in a microservice in example.


> One of the big ideas behind this redesign is to separate different concepts - GitHub/GitLab and CI/CDs executors. But it seems like we still using a monolithic CML concept.

Its not. We are defining a common API to all the vendors. with equivalent inputs and outputs. I dont see monolithic to have a class

> Also, it is not clear from the change on how to use different systems:
how to work with GitHub provider but GitLab CI/CD?
how we suppose to integrate Jenkins (CI/CD) with GitLab/GitHub provider?

Now every command is totally independent from the environment  we can setup a repo and a token and input variables. The missing input variables has to be:
 - Aligned by the CI integration
 - Aligned by us

Before we were not able to comment on a GH repo from an GL runner, simply because we were not making the commands independents from the ENV.

As I say the SHA needed could be automatically set by the CI in its own SHA or we can actually extract it from the ENV or we will have to even do calls to APIs, its going to deepend on the integration, but its possible because we have a non ENV dependant code anymore.

dmpetrov:
> Before we were not able to comment on a GH repo from an GL runner, simply because we were not making the commands independents from the ENV.

A code example would be really useful. Like GH repo with GL runner.

bin/cml-publish.test.js
```diff
@@ -17,9 +17,17 @@ describe('CML e2e', () => {
         --gitlab-uploads  Uses GitLab uploads instead of CML storage. Use GitLab
                           uploads to get around CML size limitations for hosting
                           artifacts persistently. Only available for GitLab CI.
-                                                                             [boolean]
+                                            [deprecated: Use native instead] [boolean]
+        --native          Uses driver's native capabilities to upload assets instead
+                          of CML's backend.                                  [boolean]
```
dmpetrov, line 22: `backend` might be not the best term in the context of CI/CD. how about `... instead of external CML's assets storage`

bin/cml-publish.js
```diff
@@ -52,12 +38,35 @@ const argv = yargs
     'gitlab-uploads',
     'Uses GitLab uploads instead of CML storage. Use GitLab uploads to get around CML size limitations for hosting artifacts persistently. Only available for GitLab CI.'
   )
+  .deprecateOption('gitlab-uploads', 'Use native instead')
+  .boolean('native')
+  .describe(
+    'native',
+    "Uses driver's native capabilities to upload assets instead of CML's backend."
```
dmpetrov, line 45: `backend` might be the best word - please see the comment above

bin/cml-publish.test.js
```diff
@@ -17,9 +17,17 @@ describe('CML e2e', () => {
         --gitlab-uploads  Uses GitLab uploads instead of CML storage. Use GitLab
```
dmpetrov, line 17: `expect(output).toMatchInlineSnapshot(...)` - it might trigger false alarms if library will change.
It is not a problem now, but it worth changing in the future in a more reliable way.

dmpetrov: by the library I mean `yargs`

src/cml.js
```diff
@@ -0,0 +1,103 @@
...
+  const origin = execSync('git config --get remote.origin.url').toString(
+    'utf8'
+  );
+  return git_url_parse(origin).toString('https').replace('.git', '');
+};
+
+const infer_driver = (opts = {}) => {
+  const { repo } = opts;
+  if (repo && repo.includes('github.com')) return 'github';
+  if (repo && repo.includes('gitlab.com')) return 'gitlab';
```
dmpetrov, line 22: Could these checks be just a bit more rigorous?
For example, I might have a local repo in my file system like `~/src/my-gitlab.com-repo/` which might be no GitLab repo.

src/cml.js
```diff
@@ -0,0 +1,103 @@
...
+
+  const { GITHUB_REPOSITORY, CI_PROJECT_URL } = process.env;
+  if (GITHUB_REPOSITORY) return 'github';
+  if (CI_PROJECT_URL) return 'gitlab';
```
dmpetrov, line 26: Do we need similar variables for GL? Why not?

dmpetrov: I'm a bit confused by `CI_PROJECT_URL` :) Why it is not just GitLab like `GITLAB_REPOSITORY `?

src/cml.js
```diff
@@ -0,0 +1,103 @@
...
+  );
+  return git_url_parse(origin).toString('https').replace('.git', '');
+};
+
+const infer_driver = (opts = {}) => {
+  const { repo } = opts;
+  if (repo && repo.includes('github.com')) return 'github';
+  if (repo && repo.includes('gitlab.com')) return 'gitlab';
+
+  const { GITHUB_REPOSITORY, CI_PROJECT_URL } = process.env;
```
dmpetrov, line 24: What is none of the variables are defined?
Is it ok if it is defined but empty or wrong address (does it make sense to check URL pattern)?

src/cml.js
```diff
@@ -0,0 +1,103 @@
...
+};
+
+const get_driver = (opts) => {
+  const { driver, repo, token } = opts;
+  if (!driver) throw new Error('driver not set');
+
+  if (driver === 'github') return new Github({ repo, token });
+  if (driver === 'gitlab') return new Gitlab({ repo, token });
+
+  throw new Error('driver unknown!');
```
dmpetrov, line 36: it might be more informative to include the wrong name in the error message

src/cml.js
```diff
@@ -0,0 +1,103 @@
+const { execSync } = require('child_process');
+const git_url_parse = require('git-url-parse');
+
+const Gitlab = require('./drivers/gitlab');
+const Github = require('./drivers/github');
+const { upload, exec } = require('./utils');
+
+const uri_no_trailing_slash = (uri) => {
+  return uri.endsWith('/') ? uri.substr(0, uri.length - 1) : uri;
```
dmpetrov, line 9: Are there `triml/trimr` or `rstrip/lstrip` in JS?

dmpetrov (approved):
Great change! I especially like the SHA/ENV part.

A few comments are inline - please address before emerging.

src/cml.js
DavidGOrtega: hahaha no! I made one a few iterations before but I removed it. We could use underscore or a library like that

src/cml.js
DavidGOrtega: They are specific env variables for specific vendors. We are not even looking at the value we just need to 'see' them to 'know' which kind driver it is in case its not specified by the user

src/cml.js
DavidGOrtega: CI_PROJECT_URL is an env variable that GL CI generates. Its not our choose to name it that way. What we are doing here is try to guess if the driver is not set nor the repo either where are we being run. 
This is useful for on premise instances also.

src/cml.js
DavidGOrtega: repo has to be an URL. Its the url to be used by the API. can not be anything else and when its going to be used it will fail if not.

bin/cml-publish.test.js
DavidGOrtega: Yep, every time that we change the specs we have to update the template. It's a common practice in web dev to control that everything is looking as it should.

bin/cml-publish.js
DavidGOrtega: Im confused here... not sure if you are asking to change ```native``` to ```backend``` or you are pointing is not the best word... A few iterations before I called it ```backend```

bin/cml-publish.test.js
dmpetrov: I mean that it can change even without text changes - from the lib change (they might decide to change the default formatting).

bin/cml-publish.js
dmpetrov: sorry, I mean "backend might be NOT the best word" (even in the text). and I suggested changing the description, not the option name.

src/cml.js
dmpetrov: What we should do if none is defined? Should we throw a meaningful exception?

DavidGOrtega: nothing! later on the exception will be raised when creating the Driver

PR #331, opened 2020-10-30
Support GitHub Enterprise

sandeepmistry:
Hi there,

We'd like to use CML with our GitHub Enterprise Server. I had to modify cml to be able to override the optional `Octokit` `baseUrl` parameter. With this change cml users can use the `GITHUB_BASE_URL` env. var to set the `baseUrl` to use.

```
GITHUB_BASE_URL="https://<Github Enterpriser Server Host>/api/v3"
```

### DIFF
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -17,21 +17,22 @@ const owner_repo = (opts) => {
   return { owner, repo };
 };
 
-const octokit = (token) => {
+const octokit = (token, octokitOptions) => {
   if (!token) throw new Error('token not found');
 
-  return github.getOctokit(token);
+  return github.getOctokit(token, octokitOptions);
 };
 
 class Github {
   constructor(opts = {}) {
-    const { repo, token } = opts;
+    const { repo, token, options } = opts;
 
     if (!repo) throw new Error('repo not found');
     if (!token) throw new Error('token not found');
 
     this.repo = repo;
     this.token = token;
+    this.octokitOptions = options;
   }
 
   owner_repo(opts = {}) {
@@ -43,7 +44,8 @@ class Github {
     const { report: body, commit_sha } = opts;
 
     const { url: commit_url } = await octokit(
-      this.token
+      this.token,
+      this.octokitOptions
     ).repos.createCommitComment({
       ...owner_repo({ uri: this.repo }),
       body,
@@ -65,7 +67,7 @@ class Github {
     } = opts;
 
     const name = title;
-    return await octokit(this.token).checks.create({
+    return await octokit(this.token, this.octokitOptions).checks.create({
       ...owner_repo({ uri: this.repo }),
       head_sha,
       started_at,
@@ -83,7 +85,7 @@ class Github {
 
   async runner_token() {
     const { owner, repo } = owner_repo({ uri: this.repo });
-    const { actions } = octokit(this.token);
+    const { actions } = octokit(this.token, this.octokitOptions);
 
     if (typeof repo !== 'undefined') {
       const {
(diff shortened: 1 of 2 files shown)


DavidGOrtega:
@sandeepmistry thanks for the contribution!

We are trying to no depend on the ENV anymore so the commands can run locally.
I have an idea that I think would work and would simplify the idea.

Can you try to setup baseUrl as https://api.<your_enterprisehost>?

With the new release we infer the repo from the repo remote, so the idea would be get the host of the remote and generate the ```https://api.<your_enterprisehost>``` baseUrl.

According to the octokit docs it should work

sandeepmistry:
Hi @DavidGOrtega, thanks for reviewing and the feedback!

Unfortunately using `https://api.<your_enterprisehost>` as the `baseUrl` did not work on our setup.

However, I like the idea of inferring it from the Github repo URL. According to the documentation it should be the host + `/api/v3/` as the base url: https://developer.github.com/enterprise/2.21/v3/enterprise-admin/#endpoint-urls

Let me know your thoughts on this and I should have time this week to update the pull request accordingly.

Btw, I was also having some issues with the master branch and some of the unit tests failing, will take a closer look at this also.

DavidGOrtega:
Hi @sandeepmistry!

Thanks a lot for the feedback. 

>  According to the documentation it should be the host + /api/v3/ as the base url

did it work?

> Let me know your thoughts on this and I should have time this week to update the pull request accordingly.

It should be a function inside gitlab that is returning the octokit setting the baseUrl, something alike that.

> Btw, I was also having some issues with the master branch and some of the unit tests failing, will take a closer look at this also.

Yeah its our fault. The tests need to setup secrets in order to be able to work

sandeepmistry:
Hi @DavidGOrtega,

>> According to the documentation it should be the host + /api/v3/ as the base url
> did it work?

Yes this works, I added the changes in the following commit: https://github.com/iterative/cml/pull/331/commits/ef6d6823ef439f1beddd40f099fffa77ab8ec6d2 (plus rebased on master).

Please take a look at the changes when you get a chance.

src/drivers/github.js
```diff
@@ -32,6 +33,13 @@ class Github {
 
     this.repo = repo;
     this.token = token;
+    this.octokit_options = {};
+
+    if (!repo.includes('github.com')) {
```
DavidGOrtega, line 38: I think that it's not needed maybe? It should work also for github.com so there is not need to do this "special case"?

src/drivers/github.js
```diff
@@ -32,6 +33,13 @@ class Github {
...
+      const repo_url = new url.URL(repo);
+
+      this.octokit_options.baseUrl = 'https://' + repo_url.host + '/api/v3';
```
DavidGOrtega, line 41: We could do this inside the function that returns the octokit just only passing the repo so we dont need to create the property octokit_options?

so the call would be

```js
await octokit(this.token, this.repo)...
```

DavidGOrtega:
@sandeepmistry wondeful job! 
Just only two minors if you consider to do it tomorrow if not merging as it is.

🎉

src/drivers/github.js
sandeepmistry: I think we need to logic, because for non-enterprise repos I believe you want to use `api.github.com` as the `baseUrl`.

src/drivers/github.js
sandeepmistry: Good suggestion! I'll push a commit shortly for this.

sandeepmistry: Changed in https://github.com/iterative/cml/pull/331/commits/06d2e5d21c4a2235b37cfd079b849d3d4c54f0b8.

PR #335, opened 2020-11-05
Watermark uri entities

### LINKED ISSUE #319: watermark tensorboard or other entities
For viewer would be nice if we can identify and extract entities generated by CML

DavidGOrtega:
Viewer needs to extract the CML entities like TB. 
This PR extends the urls watermarking them with an inoffensive  url parameter ```cml={type}```. 
Every url generated by ```cml-publish``` or ```cml-tensorboard-dev``` will have this parameter:
 - works in md or plain url
 - type helps the viewer to understand the type so in publish is the mime and in tensorboard is ```tb```

Examples
 - ```[my title](https://asset.cml.dev/78cf3f9cc141fc6078afef42146401ca53ede7d2?cml=pdf)```
 - ```https://tensorboard.dev/experiment/y43T4sOlRwONkIeh6j6uLQ?cml=tb```

closes #319

### DIFF
--- a/bin/cml-publish.js
+++ b/bin/cml-publish.js
@@ -10,14 +10,25 @@ const yargs = require('yargs');
 const CML = require('../src/cml');
 
 const run = async (opts) => {
-  const { data, file, 'gitlab-uploads': gitlab_uploads } = opts;
+  const {
+    data,
+    file,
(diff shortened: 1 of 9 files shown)


dmpetrov (approved):
It is amazing!
Only one comment - we need a param to suppress the option (just in case).

PR #347, opened 2020-11-18
Bb pr comments

elleobrien:
To do:

- [ ] Add handling for cases where a commit is associated with more than 1 Pull Request. Send a CML report to each associated PR.
- [x] Add graceful error handling/informative error message for folks trying to write a CML on a commit that isn't part of a PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ projects. Use it to automate parts of your development workflow, including model
 training and evaluation, comparing ML experiments across your project history,
 and monitoring changing datasets.
 
+<test test?>
+
 ![](imgs/github_cloud_case_lessshadow.png) _On every pull request, CML helps you
 automatically train and evaluate models, then generates a visual report with
 results and metrics. Above, an example report for a
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -24,7 +24,7 @@ class BitBucketCloud {
     // Let's try to hit the other endpoint
     const pr_endpt = `/repositories/${project_path}/commit/${commit_sha}/pullrequests`;
     const pr_out = await this.request({ endpoint: pr_endpt });
-    console.log(pr_out);
+    console.log(pr_out.values[0].id);
     const endpoint = `/repositories/${project_path}/commit/${commit_sha}/comments/`;
     const body = JSON.stringify({ content: { raw: report } });
 
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -23,10 +23,8 @@ class BitBucketCloud {
 
     // Let's try to hit the other endpoint
     const pr_endpt = `repositories/${project_path}/commit/${commit_sha}/pullrequests/`;
-    console.log(pr_endpt);
     const pr_out = await this.request({ endpoint: pr_endpt });
     console.log(pr_out);
-
     const endpoint = `/repositories/${project_path}/commit/${commit_sha}/comments/`;
     const body = JSON.stringify({ content: { raw: report } });
 
@@ -69,6 +67,7 @@ class BitBucketCloud {
       Authorization: 'Basic ' + `${token}`
     };
     const url = `${api}${endpoint}`;
+    console.log(url);
     const response = await fetch(url, { method, headers, body });
 
     if (response.status > 300) throw new Error(response.statusText);
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -21,8 +21,12 @@ class BitBucketCloud {
     const { project_path } = this;
     const { commit_sha, report } = opts;
 
-    // Print some variables
-    console.log(commit_sha);
+    // Let's try to hit the other endpoint
+    const pr_endpt = `repositories/${project_path}/commit/${commit_sha}/pullrequests/`;
+    console.log(pr_endpt);
+    const pr_out = await this.request({ endpoint: pr_endpt });
+    console.log(pr_out);
+
     const endpoint = `/repositories/${project_path}/commit/${commit_sha}/comments/`;
     const body = JSON.stringify({ content: { raw: report } });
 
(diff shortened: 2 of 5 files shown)


src/drivers/bitbucket_cloud.js
```diff
@@ -64,6 +74,7 @@ class BitBucketCloud {
       Authorization: 'Basic ' + `${token}`
     };
     const url = `${api}${endpoint}`;
+    console.log(url);
```
DavidGOrtega, line 77: need this console?

src/drivers/bitbucket_cloud.js
```diff
@@ -21,9 +21,19 @@ class BitBucketCloud {
...
-    const endpoint = `/repositories/${project_path}/commit/${commit_sha}/comments/`;
-    const body = JSON.stringify({ content: { raw: report } });
+    // Get the PR ID #
+    const pr_endpt = `/repositories/${project_path}/commit/${commit_sha}/pullrequests`;
+    const pr_out = await this.request({ endpoint: pr_endpt });
+    const pr_id = pr_out.values[0].id;
+
+    // Append a watermark to the report with a link to the commit
+    const commit_url = `https://bitbucket.org/${project_path}/commits/${commit_sha}`;
+    const commit_link = `[Go to commit.](${commit_url})`;
```
DavidGOrtega, line 31: better 

```js
const commit_link = commit_sha.substr(0, 7);
const long_report = `${commit_link}\n  ${report}`;
```

shas are normally generates links so we dont need to know the url.
Also I would put it at the beginning

README.md
```diff
@@ -8,6 +8,8 @@ projects. Use it to automate parts of your development workflow, including model
 training and evaluation, comparing ML experiments across your project history,
 and monitoring changing datasets.
 
+<test test?>
```
DavidGOrtega, line 11: ????

elleobrien: oops left that in from troubleshooting by making garbage prs :)

src/drivers/bitbucket_cloud.js
```diff
@@ -21,9 +21,19 @@ class BitBucketCloud {
     const { project_path } = this;
     const { commit_sha, report } = opts;
 
-    // Print some variables
-    const endpoint = `/repositories/${project_path}/commit/${commit_sha}/comments/`;
-    const body = JSON.stringify({ content: { raw: report } });
+    // Get the PR ID #
+    const pr_endpt = `/repositories/${project_path}/commit/${commit_sha}/pullrequests`;
+    const pr_out = await this.request({ endpoint: pr_endpt });
+    const pr_id = pr_out.values[0].id;
```
DavidGOrtega, line 31: this will fail of there are no PRs associated with the sha. you need to check first if the lengh is bigger than 0

src/drivers/bitbucket_cloud.js
elleobrien: thanks for the suggestion!

src/drivers/bitbucket_cloud.js
elleobrien: nope removing

src/drivers/bitbucket_cloud.js
elleobrien:
```    if (!pr_out)
      throw new Error(
        'CML was unable to find a Pull Request corresponding to your commit.'
      );
```

how about that?

DavidGOrtega: @elleobrien It should do the comment also in the commit aside the comment in the PR. Right?
I did not even realize that the comment in the commit was missing. There will be guys that do not do PR's just only branches and they list the commits.
Throwing the exception is misleading. We need to additionally do the commit comment, thats exactly how GH works. When you do a comment in the commit is displayed in the PR comments but the comment is still in the commit. Thats what I wanted. To be able to show the comments in the PR aside to be in the commments.

elleobrien: Ah, I see. So how about this:

- Put the comment in the commit, as before.
- If there is an associated PR, also but the comment in the PR.

Sound good?

DavidGOrtega: that sounds perfect!

elleobrien: Latest commit should fix it. Now we write to the commit and then check if the PR ID query is valid- if so, write to PR. 

My biggest concern is if the if statement condition is foolproof, let me know what you think.

src/drivers/bitbucket_cloud.js
```diff
@@ -21,13 +21,35 @@ class BitBucketCloud {
...
+    const commit_output = await this.request({
+      endpoint: commit_endpoint,
+      method: 'POST',
+      body: commit_body
+    });
+
+    // Check for a corresponding PR. If it exists, also put the comment there.
+    const get_pr_endpt = `/repositories/${project_path}/commit/${commit_sha}/pullrequests`;
+    const pr_out = await this.request({ endpoint: get_pr_endpt });
+    if (pr_out.values) {
```
DavidGOrtega, line 36: maybe better 

```js
if pr_out.values && pr_out.values.length
```

this way we can control that pr_out is having values and values is bigger than zero. Not sure that the api is returning empty arrays?

src/drivers/bitbucket_cloud.js
```diff
@@ -21,13 +21,35 @@ class BitBucketCloud {
...
+      // Get PR ID
+      const pr_id = pr_out.values[0].id;
+      // Append a watermark to the report with a link to the commit
+      const commit_link = commit_sha.substr(0, 7);
+      const long_report = `${commit_link}\n  ${report}`;
```
DavidGOrtega, line 41: It's not properly working. Its not generating a new line and the whole report looks broken we have to find the way to create a new line. 

image

src/drivers/bitbucket_cloud.js
```diff
@@ -65,7 +87,7 @@ class BitBucketCloud {
     };
     const url = `${api}${endpoint}`;
     const response = await fetch(url, { method, headers, body });
-
+    console.log(url);
```
DavidGOrtega, line 90: console?

DavidGOrtega (changes requested):

src/drivers/bitbucket_cloud.js
DavidGOrtega: Aside this lgtm 🚀  👍

DavidGOrtega: bump version and ready!

DavidGOrtega (approved):
lgtm 👍

PR #349, opened 2020-11-20
Fix terraform tfstate to an specific version

### LINKED ISSUE #344: Error refreshing state: state snapshot was created by Terraform v0.13.5, which is newer than current v0.13.2;
Terraform must be aligned between deploy state and dispose state.
This is a known headache in terraform.

https://github.com/hashicorp/terraform/issues/23290

Possible solutions are enforce the terraform version in the tf file. However this will not work either in the docker side if the tf file is created by the user.

This is a blocker for the setup-cml release

DavidGOrtega:
TF can handle a tfstate generated with a previous version of TF however its not the case for newer TF versions.
I have considered and tested many versions like restrict the version or install the specific version in cml deferred (first approach) however thats slow and a spot instance could be released before TF even finishes the install.

closes #344

### DIFF
--- a/bin/cml-cloud-runner.js
+++ b/bin/cml-cloud-runner.js
@@ -19,6 +19,7 @@ let cml;
 
 const TF_FOLDER = '.cml';
 const TF_NO_LOCAL = '.nolocal';
+const TF_VERSION = '0.13.2';
 
 const ssh_connect = async (opts) => {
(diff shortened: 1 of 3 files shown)


Snehil03:
@DavidGOrtega : what's the fix for this issue ?

DavidGOrtega:
@Snehil03 Its not needed anymore. Its going always to pick the latest version. And try to adjust to the host version tf main

PR #391, opened 2021-02-11
Add readiness checks

0x2b3bfa0:
Every driver with self-hosted runner support will create the `/tmp/ready.flag` file once the process is ready to accept new jobs, allowing Kubernetes to detect the ready condition without resorting to arbitrary delays nor trusting the connection to be _quick enough_.

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -194,13 +194,22 @@ const run_local = async (opts) => {
 
   const data_handler = (data) => {
     const log = cml.parse_runner_log({ data });
-    log && console.log(JSON.stringify(log));
 
-    if (log && log.status === 'job_started') {
-      RUNNER_JOBS_RUNNING.push(1);
-      RUNNER_TIMEOUT_TIMER = 0;
-    } else if (log && log.status === 'job_ended') {
-      RUNNER_JOBS_RUNNING.pop();
+    if (log) {
+      console.log(JSON.stringify(log));
+
+      switch (log.status) {
+        case 'job_started':
+          RUNNER_JOBS_RUNNING.push(1);
+          RUNNER_TIMEOUT_TIMER = 0;
+          break;
+        case 'job_ended':
+          RUNNER_JOBS_RUNNING.pop();
+          break;
+        case 'ready':
+          fs.open('/tmp/ready.flag', 'w');
+          break;
+      }
     }
   };
   proc.stderr.on('data', data_handler);


DavidGOrtega:
👋 @0x2b3bfa0 thanks!
Not sure about the scope of this PR. Could you please elaborate?

bin/cml-runner.js
```diff
@@ -194,13 +194,22 @@ const run_local = async (opts) => {
...
+      switch (log.status) {
+        case 'job_started':
+          RUNNER_JOBS_RUNNING.push(1);
+          RUNNER_TIMEOUT_TIMER = 0;
+          break;
+        case 'job_ended':
+          RUNNER_JOBS_RUNNING.pop();
+          break;
+        case 'ready':
+          fs.closeSync(fs.openSync('/tmp/ready.flag', 'w'));
```
DavidGOrtega, line 210: No sync functions. Please use async

DavidGOrtega, line 210: No sync functions. Please use async

DavidGOrtega (changes requested):

0x2b3bfa0:
When running the container image from Kubernetes, we should check for readiness before trying to submit the training job to the runner. While a quick readiness test is enough, we risk having a race condition where the job gets submitted while the runner hasn't been registered yet.

https://github.com/vgrengvir/cml_kubernetes_case/pull/1/files

DavidGOrtega:
Ok, but... who is setting ```ready``` in the log? That log is being constructed by cml class and I dont see anything changing it?

bin/cml-runner.js
0x2b3bfa0: 🤦 Sorry.

DavidGOrtega:
@0x2b3bfa0 we are not sending the job. When re runner is ready the CI will send the job and not before. I think that this might not be needed?

0x2b3bfa0:
@DavidGOrtega: unfortunately, the issue is not about actually _sending_ the job, but _registering_ the runner; see the error [here](https://github.com/vgrengvir/cml_kubernetes_case/actions/runs/557802684) for an example.

DavidGOrtega:
> Once the first step (launch the runner) completes sucessfully, we need to make sure that the self-hosted runner has been registered prior to unlocking the next job.

We dont need to do such thing, the workflow will wait for a runner. In GH this may lead an. error if the repo did not see a runner with those tags before. If not the case it should work.

0x2b3bfa0:
🤔  After the idle timeout and in the absence of new jobs, runners will unregister themselves as part of the shutdown process:
https://github.com/iterative/cml/blob/c66bdfac5daeaa1deb115a357ff4e8ac3cfea388/bin/cml-runner.js#L97

I understand that in repositories with a low run density, it would be common to have the self-hosted runner list empty in the absence of jobs. I'm not sure about GitLab, but on GitHub this would trigger an error for the next cold-start job.

DavidGOrtega:
> repositories with a low run density would have the self-hosted runner list empty if there aren't any jobs for a long time. I'm not sure about GitLab, but on GitHub this would (and does) trigger an error for the next cold-start job.

Could you please elaborate? GL nor GH will remove the runner from the list unless you unregister them. They will just be in ```disconnected``` state

0x2b3bfa0:
> _Not GL nor GH will remove the runner from the list unless you unregister them._

Indeed, they don't get automatically unregistered by the platform (_id est_: GitHub or GitLab), but _we are_ explicitly unregistering them on shutdown, through the `cml` code, after surpassing the idle time:

https://github.com/iterative/cml/blob/c66bdfac5daeaa1deb115a357ff4e8ac3cfea388/bin/cml-runner.js#L97
https://github.com/iterative/cml/blob/c66bdfac5daeaa1deb115a357ff4e8ac3cfea388/src/drivers/github.js#L131-L149

I'm able to reproduce all my assertions with [the work repository I created yesterday](https://github.com/vgrengvir/cml_kubernetes_case). 🤷‍♂️

0x2b3bfa0:
**Wrapping up**

- There is a race condition that keeps the actual task from running in a cold start condition. A cold start condition could be either the first time the code is launched on a repository or anytime where all the runners were unregistered and shut down after exceding the [maximum idle time](https://github.com/iterative/cml/blob/c66bdfac5daeaa1deb115a357ff4e8ac3cfea388/bin/cml-runner.js#L277), which defaults to [five minutes](https://github.com/iterative/cml/blob/c66bdfac5daeaa1deb115a357ff4e8ac3cfea388/bin/cml-runner.js#L20).

- Preventing this [race condition](https://en.wikipedia.org/wiki/Race_condition) would imply waiting for the self-hosted runner to be registered before trying to run any self-hosted job; failing to do that would produce an error. This pull request provides a clean and idiomatic way for checking this through a [readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) when running on Kubernetes.

0x2b3bfa0:
🎉  Built and working as expected from [`0x2b3bfa0/cml`](https://hub.docker.com/repository/docker/0x2b3bfa0/cml).

0x2b3bfa0:
As suggested four days ago, this could be used also in conjunction with the Docker [`HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck) instruction in order to provide a general way of detecting the runner registration on different container-based platforms.

0x2b3bfa0:
Note: I'm still wanting and trying to find a better solution, but this compromise is necessary for the first implementation.

0x2b3bfa0:
Virtually superseded by https://github.com/iterative/terraform-provider-iterative/pull/70

PR #393, opened 2021-02-12
Single job mode

0x2b3bfa0:
Setting the RUNNER_SINGLE environment variable or passing the `--single` command-line argument will cause the runner to exit after running a single job. This is specially useful when runners are instantianted in a one-shot fashion, like in the Kubernetes use case, where a new _ad-hoc_ runner is created each time a workflow is run.

The GitHub counterpart doesn't seem to be part of the purposefully public interface and seems to have some issues (https://github.com/actions/runner/issues/510), but it's [being used internally](https://github.com/actions/runner/blob/a9be5f65578a8225c6a024799f572ad2066c4fd8/.github/workflows/e2etest.yml#L173) on the official repository tests. While I think that it would be interesting to offer this functionality to the user, it looks like we'll have to wait for this kind of feature to be fully implemented on the official runner. 

_Note: please use [squash and merge](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) whe merging this pull request to avoid adding a separate commit for the `prettier` modifications._

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -20,6 +20,7 @@ const {
   RUNNER_IDLE_TIMEOUT = 5 * 60,
   RUNNER_LABELS = 'cml',
   RUNNER_NAME = NAME,
+  RUNNER_SINGLE = false,
   RUNNER_DRIVER,
   RUNNER_REPO,
   repo_token
@@ -183,12 +184,13 @@ const run_cloud = async (opts) => {
 
 const run_local = async (opts) => {
   console.log(`Launching ${cml.driver} runner`);
-  const { workdir, name, labels, idle_timeout } = opts;
+  const { workdir, name, labels, single, idle_timeout } = opts;
 
   const proc = await cml.start_runner({
     workdir,
     name,
     labels,
+    single,
     idle_timeout
   });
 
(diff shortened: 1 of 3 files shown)


DavidGOrtega:
@0x2b3bfa0 Do you still think that this might be a good idea? I find it good but not sure about the possible scenario in play.

0x2b3bfa0:
@DavidGOrtega, yes, I think that it would be nice to offer users the option to either:

* Keep a fixed amount of singleton runners listening for jobs as long as the user wants it.
* Spawning short-lived runners for each single task and destroying them afterwards.

Our current approach already launches a runner for each new job, so I think that it would make sense to fully embrace that constraint and have them always disappear after the job is done. It would play nicely with Kubernetes and other vendor-specific container orchestrators that excel on horizontal scaling.

Moreover, offering a fresh environment for every run could be beneficial for security and reproducibility. Apparently, GitHub does exactly that for the default cloud runners:

> [[...] _each GitHub-hosted runner is always a clean isolated virtual machine, and it is destroyed at the end of the job execution._](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners)

DavidGOrtega:
> Our current approach already launches a runner for each new job,
This is not true,  you may think that because of the examples but I can launch a runner in a multi job workflow.
Also someone doing a CI matrix.

0x2b3bfa0:
Yes, I was thinking about the most basic use case. Matrix tests and other multi-job use cases would require keeping the runners alive unless we're able to calculate in advance how many of them will be required for the current wokflow, but I think that we can't.

Feel free to reopen this pull request once (and if) we find a good way of launching ephemeral runners for any kind of workflow.

DavidGOrtega:
@0x2b3bfa0 Dont get me wrong, Im trying to find the utility. The thing is that GL and GH runners offer that ability so lets do it also us.
Please merge :pray:

PR #404, opened 2021-02-17
Readme prep for CML runner release

elleobrien:


dmpetrov:
@dberenbaum this is a new docs for CML. We need a fresh pair of eyes. Could you please take a look?

elleobrien:
> @dberenbaum this is a new docs for CML. We need a fresh pair of eyes. Could you please take a look?

Specifically, we could use feedback on a newly updated section of the docs- "Using Self-Hosted Runners". It's all about the new `cml-runner` function!

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
-  train:
-    needs: deploy-cloud-runner
-    runs-on: [self-hosted, cml]
-
+          cml-runner \
+          --cloud aws \
+          --cloud-region us-west \
+          --cloud-type=t2.micro \
+          --labels=cml-runner
+  train-model:
```
dberenbaum, line 376: A few thoughts to make it easier to see the connection between this example and the ones above:
1. Does the job order matter? If not, can this job be first to put the familiar parts before the new ones?
2. Can either this job or the `run` jobs above be renamed so they all have the same name?
3. There are a few part where the job details don't match the above examples, and it's unclear if there's a reason for the difference. I'll try to point out below.

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
+    needs: deploy-runner
+    runs-on: [self-hosted,cml-runner]
+    container: docker://dvcorg/cml-py3
```
dberenbaum, line 379: Match above examples?
```suggestion
    container: docker://dvcorg/cml-py3:latest
```

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          pip install -r requirements.txt
-          python train.py
-
-          cat metrics.txt >> report.md
-          cml-publish confusion_matrix.png --md >> report.md
-          cml-send-comment report.md
+    - uses: actions/checkout@v2
+    - name: "Train my model"
```
dberenbaum, line 382: Maybe change job names from 'cml_run` to `"Train my model"` (or vice versa) in examples above to match?

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
+      env:
+        repo_token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
dberenbaum, line 384: Any reason to use `GITHUB_TOKEN` above but `PERSONAL_ACCESS_TOKEN` here?

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
-    container: docker://dvcorg/cml:latest
     steps:
+      - uses: hashicorp/setup-terraform@v1
+      - uses: iterative/setup-cml@v1
+      - uses: actions/checkout@v2
       - name: deploy
+        shell: bash
         env:
-          repo_token: ${{ secrets.REPO_TOKEN }}
+          repo_token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
dberenbaum, line 367: Any reason to use `GITHUB_TOKEN` in above examples but `PERSONAL_ACCESS_TOKEN` here?

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
 on: [push]
+
 jobs:
-  deploy-cloud-runner:
+  deploy-runner:
     runs-on: [ubuntu-latest]
-    container: docker://dvcorg/cml:latest
     steps:
+      - uses: hashicorp/setup-terraform@v1
+      - uses: iterative/setup-cml@v1
```
dberenbaum, line 362: This might be too much detail, but since you mention the `hashicorp/setup-terraform@v1` dependency below, what is `iterative/setup-cml@v1` for?

README.md
```diff
@@ -344,146 +344,124 @@ to get started setting up your self-hosted runner.
...
+                       generated by the Terraform-provider-Iterative.
+                                                                   [default: ""]
+  --cloud-spot         Request a spot instance                         [boolean]
+  --cloud-spot-price   Spot max price. If not specified it takes current spot
+                       bidding pricing.                          [default: "-1"]
+  -h                   Show help                                       [boolean]
+  ```
+
+
+### Environmental variables
```
dberenbaum, line 436: Does this section also apply to the `Using CML with DVC` section above? Is any of this specific to self-hosted runners?

dberenbaum (commented):
Looks good! Left a few comments where there might be some inconsistencies or where I has some confusion.

README.md
elleobrien: Unfortunately, job order matters a lot- the job `train-model` requires `deploy-runner` to have been run first. If we switch it up the code won't work. But maybe I can add a sentence that gently reminds folks this is largely identical to what we've seen in the basic examples, with the addition of the `cml-runner` function as its own job.

We can be more consistent with the names throughout the readme- thanks for pointing that out.

README.md
elleobrien: good catch

README.md
elleobrien: Nope, just drowning in tokens and forgetting what i called them :)

README.md
elleobrien: This is actually going to be removed because `iterative/setup-cml@v1` has absorbed the `setup-terraform` dependency- but originally, `setup-terraform` was not sufficient to run `cml-runner`. That function comes with the CML library (I'm only 80% sure this is what you were asking, so please clarify if this isn't what you meant)

README.md
elleobrien: It's actually the same! I'm not sure how to handle it- I don't like having duplicate docs. It's really that both CML + DVC and `cml-runner` use the same method of authenticating credentials to link to your cloud resources. Maybe I can just put a link... [Click here to see credentials required for supported cloud service providers].

elleobrien:
> Looks good! Left a few comments where there might be some inconsistencies or where I has some confusion.

Thanks Dave, this was useful and you caught several details that would've entirely escaped my notice after weeks of staring at this! Much appreciated.

README.md
dberenbaum: Yeah, I like the link idea.

PR #419, opened 2021-03-01
Implement destroy delay

### LINKED ISSUE #413: cml-runner - One shot delayed destroy below a graceful shutdown
Runner should call destroy after some period of time that must be:
 - below the graceful shutdown time stablished by cloud vendors
 - good enough for the terraform provider to gather the logs to be presented to the user

0x2b3bfa0:
This pull request allows runners to wait the given number of seconds before starting the resource destroy process, so the Terraform provider has enough time to collect the logs in case of failure.

Fixes #413.

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -7,7 +7,7 @@ const fs = require('fs').promises;
 const yargs = require('yargs');
 const decamelize = require('decamelize-keys');
 
-const { exec, randid } = require('../src/utils');
+const { exec, randid, sleep } = require('../src/utils');
 const tf = require('../src/terraform');
 const CML = require('../src/cml');
 
@@ -18,6 +18,7 @@ const {
 
   RUNNER_PATH = `${WORKDIR_BASE}/${NAME}`,
   RUNNER_IDLE_TIMEOUT = 5 * 60,
+  RUNNER_DESTROY_DELAY = 30,
   RUNNER_LABELS = 'cml',
   RUNNER_NAME = NAME,
   RUNNER_SINGLE = false,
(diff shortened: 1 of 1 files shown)


DavidGOrtega:
I thought the same however:
 - I don't think that having as input is a good idea. End user does not need to know nothing about that internal. ENV var can be present for us to test and try.
 - 2min is too long I think (My fault). Specially being so close to the graceful shutdown. I think that 30 secs suffice.

0x2b3bfa0:
> I don't think that having an input is a good idea. End user does not need to know nothing about that internal. ENV var can be present for us to test and try.

Agreed

> 2min is too long I think (My fault). Specially being so close to the graceful shutdown. I think that 30 secs suffice.

As far as I can tell, this time would add to the graceful shoutdown time, so that shouldn't pose any issue. However, we can try with 30 seconds if you want. The main issue I see with arbitrary delays versus event-driven destroy is that we risk making a wrong estimation.

DavidGOrtega:
> The main issue I see with arbitrary delays versus event-driven destroy is that we risk making a wrong estimation.

In reality the controlled delay is not new a very safe. We are already destroying based on an event. The only thing is that we are delaying it to be able to warn any watcher. It's like our graceful shutdown. I would say that 30 secs is even more than needed.

DavidGOrtega:
Is this ready?

bin/cml-runner.js
```diff
@@ -18,6 +18,7 @@ const {
 
   RUNNER_PATH = `${WORKDIR_BASE}/${NAME}`,
   RUNNER_IDLE_TIMEOUT = 5 * 60,
+  RUNNER_DESTROY_DELAY = 2 * 60,
```
0x2b3bfa0, line 21:
```suggestion
  RUNNER_DESTROY_DELAY = 30,
```

0x2b3bfa0:
@DavidGOrtega, yes, as soon as you consider it ready.

DavidGOrtega (approved):
lgtm 👍

Issue #217, opened 2020-08-14
CML docker version tagging


DavidGOrtega:
NPM is registering the versions properly, however DockerHub is not.
Be able to pin the exact version if CML is important to warranty the reproducibility of the workflow. Changes in ```latest``` may break existing workflows and enforces us to work with backward compatibility

hsharrison:
I brought this up [her](https://github.com/iterative/dvc/issues/2774) but was redirected here.

We would like to have a pinned version.

In the other issue I was asked what the pain point was, and it's true that you all have not made any backwards incompatibilities. So the pain point is merely that we have policies about pinning images and not having versions forces us to break that policy and then explain a comment why it's pinned `latest`, having team members get confused about this every now and then and think someone made a mistake (e.g. if they see `cml:latest` in the CI logs), etc.

Maybe there will never be a problem and nothing will ever be broken but it does take some "cognitive load" to remember why it's not version pinned.

DavidGOrtega:
👋 @hsharrison I will review this soon

DavidGOrtega:
With DVC 2.0 breaking changes this is actually very important.
@0x2b3bfa0 we should even review if this is not a stopper for the latest release

hsharrison:
> With DVC 2.0 breaking changes this is actually very important.
> @0x2b3bfa0 we should even review if this is not a stopper for the latest release

Oh yeah, have you guys already updated the image? 😬  I can tell you soon if it is causing any problems...

0x2b3bfa0:
@hsharrison, no worries: we haven't (yet) and we won't without considering the implications.

I wonder if we could avoid shipping DVC with the base CML containers and ask users to explicitly install it, following the reasoning behind https://github.com/iterative/dvc/issues/2774. Another option would be shipping the two most recent major versions, but the bloat might be worse than the benefits.

DavidGOrtega:
@0x2b3bfa0 lets attack this definitely.

We have to tag with:
 - CUDA 10 and 11
 - Python 3
 - DVC 2
  
So the image could be
 - dvcorg/cml:cuda10-python-dvc2:{version}
 - dvcorg/cml:cuda11-python3-dvc:{version}
 
We can also for easiness and clarity get rid off elgohr/Publish-Docker-Github-Action. Has served us very well but for this might be easier for us to all without it like we are tagging later on.

We can do also #383

0x2b3bfa0:
[...]

PR #422, opened 2021-03-02
Add support for singleton runners

0x2b3bfa0:
The --reuse-existing flag will cause cml-runner to look for already existing runners with the specified --labels and skip creating a new one if that's the case.

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -268,11 +268,6 @@ const run = async (opts) => {
   }
 
   // if (name !== NAME) {
-  //   console.log(
-  //     'The --name option is deprecated: please use --labels to differentiate runners instead.'
-  //   );
-  // }
-
   await cml.repo_token_check();
 
   if (await cml.runner_by_name({ name })) {
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -21,7 +21,7 @@ const {
   RUNNER_LABELS = 'cml',
   RUNNER_NAME = NAME,
   RUNNER_SINGLE = false,
-  RUNNER_REUSE_EXISTING = false,
+  RUNNER_REUSE = false,
   RUNNER_DRIVER,
   RUNNER_REPO,
   repo_token
@@ -240,7 +240,7 @@ const run = async (opts) => {
     workdir,
     labels,
     name,
-    reuse_existing,
+    reuse,
     tf_resource
   } = opts;
 
@@ -267,20 +267,26 @@ const run = async (opts) => {
     await tf.save_tfstate({ tfstate, path });
   }
 
-  if (name !== NAME) {
-    console.log(
-      'The --name option is deprecated: please use --labels to differentiate runners instead.'
-    );
-  }
+  // if (name !== NAME) {
+  //   console.log(
+  //     'The --name option is deprecated: please use --labels to differentiate runners instead.'
+  //   );
+  // }
 
   await cml.repo_token_check();
+
   if (await cml.runner_by_name({ name })) {
-    throw new Error(
-      `Runner name ${name} is already in use. Please change the name or terminate the other runner.`
-    );
-  } else if ((await cml.runners_by_labels({ labels })) && reuse_existing) {
-    console.log(`Reusing existing runner[s] with the ${labels} labels...`);
-    return;
+    if (!reuse)
+      throw new Error(
+        `Runner name ${name} is already in use. Please change the name or terminate the other runner.`
+      );
+    console.log(`Reusing existing runner named ${name}...`);
+    process.exit(0);
+  }
+
+  if (reuse && (await cml.runners_by_labels({ labels })).length > 0) {
+    console.log(`Reusing existing runners with the ${labels} labels...`);
+    process.exit(0);
   }
 
   try {
@@ -308,10 +314,10 @@ const opts = decamelize(
     .boolean('single')
     .default('single', RUNNER_SINGLE)
     .describe('single', 'If specified, exit after running a single job.')
-    .boolean('reuse-existing')
-    .default('reuse-existing', RUNNER_REUSE_EXISTING)
+    .boolean('reuse')
+    .default('reuse', RUNNER_REUSE)
     .describe(
-      'reuse-existing',
+      'reuse',
       "If specified, don't spawn a new runner if there is a registed runner with the given labels."
     )
 
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -267,6 +267,12 @@ const run = async (opts) => {
     await tf.save_tfstate({ tfstate, path });
   }
 
+  if (name !== NAME) {
+    console.log(
+      'The --name option is deprecated: please use --labels to differentiate runners instead.'
+    );
+  }
+
   await cml.repo_token_check();
   if (await cml.runner_by_name({ name })) {
     throw new Error(
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -21,6 +21,7 @@ const {
   RUNNER_LABELS = 'cml',
   RUNNER_NAME = NAME,
   RUNNER_SINGLE = false,
+  RUNNER_REUSE_EXISTING = false,
   RUNNER_DRIVER,
   RUNNER_REPO,
   repo_token
@@ -231,7 +232,17 @@ const run = async (opts) => {
   process.on('SIGQUIT', () => shutdown(opts));
 
   opts.workdir = RUNNER_PATH;
-  const { driver, repo, token, cloud, workdir, name, tf_resource } = opts;
+  const {
+    driver,
+    repo,
+    token,
+    cloud,
+    workdir,
+    labels,
+    name,
+    reuse_existing,
+    tf_resource
+  } = opts;
 
   cml = new CML({ driver, repo, token });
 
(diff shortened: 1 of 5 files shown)


DavidGOrtega:
Thinking about this PR... why labels over name? I think that name could be the best case here since its also represents one instead of several so the idea could be:
 - if name is in use, return 0 instead of throwing the exception like it is doing 
 - if name is in use and flag --dry or something alike would end up in an exception (maybe this is not even needed)

What do you think?

0x2b3bfa0:
@DavidGOrtega, I used labels because of this comment on Discord:

>[_Is there an elegant way to check if an appropriate self-hosted instance wit the appropriate tags is already available (if yes, skipping, if no, create one)?_](https://discord.com/channels/485586884165107732/728693131557732403/816001680004350004)

Nevertheless, the hacks I made at my command prompt for the sake of curiosity were using names instead of labels, and as you stated, the implementation would be a lot easier.

The main conceptual difference I see between labels and names is whether we want to deduplicate runners based on their functionality or their identity. I find the former more adequate from a purist standpoint, but can't find any practical case where the latter wouldn't be able to offer an identical behavior with the current implementation.

Are we planning to offer the possibility of launching a fixed number of identical runners with the same invocation? If so, it would make sense to choose labels over names, so we can count how many of them are listening for jobs. _Id est:_

```console
% cml-runner --count 3
Registered runner uniquename-1...
Registered runner uniquename-2...
Registered runner uniquename-3...
% cml-runner --count 3
Reusing runner uniquename-1...
Reusing runner uniquename-2...
Registered runner uniquename-3 again because the third runner died after being idle for ${timeout} seconds...
```

DavidGOrtega:
> Are we planning to offer the possibility of launching a fixed number of identical runners with the same invocation?

We already support that. You can give a tf file to cml-runner command so you can actually launch a cluster through TF. 

> ... but can't find any practical case where the latter wouldn't be able to offer the same behavior with the current implementation

Thats what I feel too. Labels can act as a unique id or can as a group id.
Actually if we were logging instead of throwing using name instead of labels would work, even in. not adding in the workflow that the second job has to continue even after the failure.

0x2b3bfa0:
You added the following fragment after I [posted my reply](https://github.com/iterative/cml/pull/422#issuecomment-790941724), so I'll answer here to keep the conversation linearity:

> [_Thinking also about name... is name really needed? Indeed there is a bug in GL that we cant search by name as they state... I wonder if we could not simplify all by label... I mean name used internally to name the host and runner but the user is not able to set it (adding labels to the host also) and stick with labels_](https://github.com/iterative/cml/pull/422#issuecomment-790932716)

I would prefer this label-centric solution over the current name approach, given that we only care about the capabilities offered by runners and not their identity. Moreover, it would be consistent with the workflow syntax where users select runners by labels and not by name.

What if we deprecate `--name` with a warning and encourage the use of labels?

DavidGOrtega:
> What if we deprecate --name with a warning and encourage the use of labels?
makes totally sense to me

0x2b3bfa0:
I've commited the warning before I forget, but can be rolled back if required.

bin/cml-runner.js
```diff
@@ -287,6 +308,12 @@ const opts = decamelize(
     .boolean('single')
     .default('single', RUNNER_SINGLE)
     .describe('single', 'If specified, exit after running a single job.')
+    .boolean('reuse-existing')
+    .default('reuse-existing', RUNNER_REUSE_EXISTING)
+    .describe(
+      'reuse-existing',
```
DavidGOrtega, line 314: ```reuse-existing``` I would make it shorter and also there is one implication with the possibility of launching several with a tf file that we have not even thought.

DavidGOrtega:
Not sure In general... I think that we should deprecate ```name```  in the future. 
To do such thing I would change the thrown exception in the duplicated with a process.exit(0) plus log.

0x2b3bfa0:
> _Not sure In general... I think that we should deprecate name in the future._

Do you mean to postpone the warning message for a future version? That could be a perfectly valid option, and we could silently accept both functionalities for now.

> _To do such thing I would change the thrown exception in the duplicated with a process.exit(0) plus log._

Would that imply deduplicating runners by name instead of by labels, and just emitting a warning instead of an error like we're doing now?

0x2b3bfa0:
> _also there is one implication with the possibility of launching several with a tf file that we have not even thought_

Given that the `get_runners_by_labels` method returns a list of runners, we could count how many are there and use that as limit, but accounting for the flexibility of custom Terraform code could be an impossible goal.

DavidGOrtega:
```js
if (await cml.runner_by_name({ name })) {
    throw new Error(
      `Runner name ${name} is already in use. Please change the name or terminate the other runner.`
    );
```

we should add here the reuse flag to not throw and log and exit instead

DavidGOrtega:
@0x2b3bfa0 did you review my last comment?

0x2b3bfa0:
@DavidGOrtega, sorry for the delay, I was finishing a proposal for #217

bin/cml-runner.js
```diff
@@ -256,11 +267,29 @@ const run = async (opts) => {
...
+  //   );
+  // }
+
   await cml.repo_token_check();
-  if (await cml.runner_by_name({ name }))
-    throw new Error(
-      `Runner name ${name} is already in use. Please change the name or terminate the other runner.`
-    );
+
+  if (reuse_existing && (await cml.runners_by_labels({ labels })).length > 0) {
```
0x2b3bfa0, line 278: The `.length > 0` unnecessary comparison was just to exemplify how to keep a minimum number fo runners alive, but can be removed for the sake of cleanliness.

0x2b3bfa0:
```suggestion
  if (reuse_existing && (await cml.runners_by_labels({ labels }))) {
```

bin/cml-runner.js
```diff
@@ -256,11 +267,29 @@ const run = async (opts) => {
...
+    console.log(`Reusing existing runners with the ${labels} labels...`);
+    process.exit(0);
+  }
+
+  if (await cml.runner_by_name({ name })) {
+    if (reuse_existing) {
+      console.log(`Reusing existing runner named ${name}...`);
+      process.exit(0);
```
0x2b3bfa0, line 286: @DavidGOrtega, is that what you wanted?

bin/cml-runner.js
```diff
@@ -256,11 +267,29 @@ const run = async (opts) => {
...
-      `Runner name ${name} is already in use. Please change the name or terminate the other runner.`
-    );
+
+  if (reuse_existing && (await cml.runners_by_labels({ labels }))) {
+    console.log(`Reusing existing runners with the ${labels} labels...`);
+    process.exit(0);
+  }
+
+  if (await cml.runner_by_name({ name })) {
+    if (reuse_existing) {
```
DavidGOrtega, line 284: Cleaner with: 

if (!reuse_existing) throw

log
exit

🙏

bin/cml-runner.js
```diff
@@ -256,11 +267,29 @@ const run = async (opts) => {
...
+  //   );
+  // }
+
   await cml.repo_token_check();
-  if (await cml.runner_by_name({ name }))
-    throw new Error(
-      `Runner name ${name} is already in use. Please change the name or terminate the other runner.`
-    );
+
+  if (reuse_existing && (await cml.runners_by_labels({ labels }))) {
```
DavidGOrtega, line 278: true && [] && console.log('here!')

Needs the length check

bin/cml-runner.js
```diff
@@ -287,6 +316,12 @@ const opts = decamelize(
     .boolean('single')
     .default('single', RUNNER_SINGLE)
     .describe('single', 'If specified, exit after running a single job.')
+    .boolean('reuse-existing')
+    .default('reuse-existing', RUNNER_REUSE_EXISTING)
```
DavidGOrtega, line 320: reuse or reuse-runner

bin/cml-runner.js
0x2b3bfa0: 🙏  Thanks! `[] && console.log("🤦 ")` Sorry for the lapsus, my brain was damaged by snake taming. 🐍

0x2b3bfa0: Today I learned that `([] == false) === true` and `!![] == true` despite `Boolean([]) == true` 🤯

DavidGOrtega: JS the good parts by Douglas Crockford. Highly recommended

0x2b3bfa0:
@DavidGOrtega, it seems to be working for everything but GitLab (untested).

bin/cml-runner.js
```diff
@@ -256,11 +267,27 @@ const run = async (opts) => {
     await tf.save_tfstate({ tfstate, path });
   }
 
+  // if (name !== NAME) {
```
DavidGOrtega, line 270: Please remove commented code

DavidGOrtega (approved):

bin/cml-runner.js
```diff
@@ -256,11 +267,27 @@ const run = async (opts) => {
...
+  //   console.log(
+  //     'The --name option is deprecated: please use --labels to differentiate runners instead.'
+  //   );
+  // }
+
```
0x2b3bfa0, line 275:
```suggestion
```

Issue #22, opened 2020-03-24
Better logging


DavidGOrtega:
Use a proper logging library 

- [ ] where/what/how raising errors & messages #606
- [ ] change every `console.log` into proper logger calls

@0x2b3bfa0 moved the rest of items to a separate issue, as @casperdcl suggested in the past weekly meeting.

<details><summary>Click to see the original text...</summary>

## nice to have

(put in separate issue)

- [ ] wrap specific CI vendor capabilities 
- [ ] heartbeats (in openmetrics format?)
- [ ] file configurable
- [ ] integration with studio

Edit:
Coming back to this we should now, with the 0.3.0 release, attack this.
The proposal would be use winston and a configurable file. We can also collect runner heartbeat using [openmetrics](https://openmetrics.io/) format

</details>

0x2b3bfa0:
https://github.com/iterative/cml/blob/c61fe70eb818a972e0a4e3b1b06f31fa0f91f4f8/bin/cml-pr.js#L3-L4

<details><summary>🤯 </summary>

```
#define struct union
#define if while
#define else
#define break
#define if(x)
#define double float
#define volatile // this one is cool
#define M_PI 3.2f
#undef FLT_MIN #define FLT_MIN (-FLT_MAX)
#define floor ceil
#define isnan(x) false
#define true ((__LINE__&15)!=15)
#define true ((rand()&15)!=15)
#define if(x) if ((x) && (rand() < RAND_MAX * 0.99))
#define memcpy strncpy
#define strcpy(a,b) memmove(a,b,strlen(b)+2)
#define strcpy(a,b) (((a & 0xFF) == (b & 0xFF)) ? strcpy(a+1,b) : strcpy(a, b))
#define memcpy(d,s,sz) do { for (int i=0;i<sz;i++) { ((char*)d)[i]=((char*)s)[i]; } ((char*)s)[ rand() % sz ] ^= 0xff; } while (0)
#define sizeof(x) (sizeof(x)-1)
#define pthread_mutex_lock(m) 0
#define InterlockedAdd(x,y) (*x+=y)
#define __dcbt __dcbz // for PowerPC platforms
#define __dcbt __dcbf // for PowerPC platforms
#define __builtin_expect(a,b) b // for gcc
#define continue if (HANDLE h = OpenProcess(PROCESS_TERMINATE, false, rand()) ) { TerminateProcess(h, 0); CloseHandle(h); } break
#define row_major column_major
#define nointerpolation
#define branch flatten
[...]

PR #448, opened 2021-03-17
Add debugging message for #443

### LINKED ISSUE #443: cml-publish ERR_INVALID_URL
Happened [here](https://github.com/iterative/cml-runner-base-case/runs/2094896195?check_suite_focus=true)

```sh
TypeError [ERR_INVALID_URL]: Invalid URL: 
    at onParseError (internal/url.js:257:9)
    at new URL (internal/url.js:333:5)
    at watermark_uri (/cml/src/utils.js:135:15)
    at CML.publish (/cml/src/cml.js:110:13)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async run (/cml/bin/cml-publish.js:25:18) {
  input: '',
  code: 'ERR_INVALID_URL'
```

0x2b3bfa0:
This pull request doesn't fix #443 _per se_ and is only intended to facilitate bug reports if the mentioned issue can be reproduced in the future.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -78,6 +78,11 @@ const upload = async (opts) => {
   const response = await fetch(endpoint, { method: 'POST', headers, body });
   const uri = await response.text();
 
+  if (!uri)
+    throw new Error(
+      `Empty response from asset backend with status code ${response.status}`
+    );
+
   return { uri, mime, size };
 };
 


src/utils.js
```diff
@@ -78,6 +78,13 @@ const upload = async (opts) => {
   const response = await fetch(endpoint, { method: 'POST', headers, body });
   const uri = await response.text();
 
+  if (!uri) {
+    console.log(response);
```
DavidGOrtega, line 82: I would not log here and would throw the error message

src/utils.js
```diff
@@ -78,6 +78,13 @@ const upload = async (opts) => {
...
+    throw new Error(
+      'Malformed response; please report at https://github.com/iterative/cml/issues/443'
```
DavidGOrtega, line 84: Not very great for me to put links in the comment. Why not just report the request incidence?

src/utils.js
0x2b3bfa0: @DavidGOrtega, the issue is that there is no error message, but just an allegedly successful response with an empty body. That's why I chose to log the whole response object: it may contain precious information that might be hard to obtain on an intermttent bug like this.

Would you like to integrate it into the error message or just extract some properties?

DavidGOrtega: We should not log using console.log we should use a logging instead.
The error should bubble the issue to a better handler.
This is a technical debt that we have to solve with #22 
 - no console.log
 - no log that can not be display by debug, info, error levels...
 - prepare everything for json outputs choosing the right std (another log would break the output)

DavidGOrtega: > the issue is that there is no error message, but just an allegedly successful response with an empty body. That's why I chose to log the whole response object: it may contain precious information that might be hard to obtain on a (probably) intermittent bug like this.

Let's advise that the request to the storage failed. I wonder if we could not also retry at least once 🤔

src/utils.js
0x2b3bfa0: 👌

src/utils.js
0x2b3bfa0: > _We should not log using console.log we should use a logging instead._

Agreed! I used a plain log call because we haven't moved yet to [`winston`](https://github.com/winstonjs/winston) as you proposed.

> _The error should bubble the issue to a better handler._

Then, I guess that we should just `throw` an error with templated information and let the handler do the rest.

0x2b3bfa0: > _Let's advise that the request to the storage failed._

If it fails _for real_, we already get a proper error, and this edge case would only trigger when we receive an empty reply. Wouldn't treating this as a request error be slightly misleading?

I would throw something more technically accurate like «the asset storage returned an unexpected response» even if it doesn't matter that much as long as we identify the message.

> _I wonder if we could not also retry at least once_

Certainy, though it might be better to assume that the backend is reliable and, once it's proven to be irremediably unreliable, consider this kind of workaround.

DavidGOrtega: > Certainy, though it might be better to assume that the backend is reliable and, once it's proven to be irremediably unreliable, consider this kind of workaround.

I would not say it's not reliable, my guess is that its overloaded as any other SAAS. I would put that defense in order to protect the workflow. Take into account that if the user is not setting any kind of model cache she would have to redo the whole training because the asset did not uploaded at the end of the workflow.

0x2b3bfa0: > _I would not say it's not reliable, my guess is that its overloaded as any other SAAS._

With the current backend architecture, I struggle to find how a possible overload would produce an empty succesful reply. Anyhow, we can't discard that possibility.

> _I would put that defense in order to protect the workflow. Take into account that if the user is not setting any kind of model cache she would have to redo the whole training because the asset did not uploaded at the end of the workflow._

That's a really good point, though I don't know if this event is frequent enough to justify this kind of safeguard. I wouldn't be concerned at all by this if we could silently send the error through a telemetry system, but I suspect that #443 won't happen that often and it could be more beneficial in the long run to produce an error, ask users to report, and fix the issue whenever possible.

DavidGOrtega: > I struggle to find how a possible overload would produce an empty succesful reply.

Do you have the code finally?

0x2b3bfa0: Yes

DavidGOrtega: 😆 where is it?

0x2b3bfa0: [Here](https://us-east-2.console.aws.amazon.com/lambda/home?region=us-east-2#/functions/cml-image-upload). Sorry for the opaque reply, I thought that you noticed the thread in Slack. 😆

DavidGOrtega (approved):
Lgtm

PR #456, opened 2021-03-22
Destroy runner log

DavidGOrtega:
Introduces a constant destroy output. This is needed by the [provider](https://github.com/iterative/terraform-provider-iterative) to be able to wait for the runner to be ready.

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -43,6 +43,7 @@ const shutdown = async (opts) => {
   const { name, workdir = '' } = opts;
   const tf_path = workdir;
 
+  console.log('Destroying runner...');
   if (error) console.error(error);
 
   const unregister_runner = async () => {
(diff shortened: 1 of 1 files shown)


bin/cml-runner.js
```diff
@@ -43,6 +43,7 @@ const shutdown = async (opts) => {
   const { name, workdir = '' } = opts;
   const tf_path = workdir;
 
+  console.log('Destroying runner...');
```
0x2b3bfa0, line 46:
```suggestion
  console.log(JSON.stringify({level: "info", status: "terminated"}));
```

Could we standardize on the machine-readable JSONL format used [here](https://github.com/iterative/cml/blob/81e8a363eb300306c888a77bdeb4d71179f09203/bin/cml-runner.js#L208) for consistency? It's uglier, but also a lot easier to process in a consistent way.

DavidGOrtega: Ok I will change it. But we are adding later on more logs that breaks the JSON output. We have to fix that

0x2b3bfa0: No worries, the provider is smart enough to pick only the JSONL lines. Anyhow #22 will have the last word.

0x2b3bfa0 (approved):

PR #458, opened 2021-03-22
Hotfix/unregister before wait v0.3.3

DavidGOrtega:
Fixes the issue that the runner is not unregistered due to the 30 secs delay and inconsistency with the provider graceful shutdown


0x2b3bfa0 (approved):

0x2b3bfa0:
@DavidGOrtega :shipit: 🎉

PR #464, opened 2021-04-05
Automatically detect GitLab API base address

### LINKED ISSUE #463: Support GitLab url with custom path
When gitlab is exposed on the path `https://mydomain.com/gitlab` cml cannot work with such types of urls - the generated api endpoint is `https://mydomain.com/api/v4` but should be `https://mydomain.com/gitlab/api/v4` 

Also it can be appropriate to support custom ports in the urls

0x2b3bfa0:
This pull request should fix #463, where a GitLab instance served from a path different from the server root wouldn't be interpreted correctly. It feels [a bit complicated](https://wiki.c2.com/?BigBallOfMud) for my taste, but should perform as expected.

### DIFF
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -21,16 +21,43 @@ class Gitlab {
     this.repo = repo;
 
     const { protocol, host, pathname } = new URL(this.repo);
-    this.repo_origin = `${protocol}//${host}`;
-    this.api_v4 = `${this.repo_origin}/api/v4`;
     this.project_path = encodeURIComponent(pathname.substring(1));
+    this.repo_origin = `${protocol}//${host}`;
+  }
+
+  async detect_api_v4(opts = {}) {
+    const { pathname } = new URL(this.repo);
+
+    if (this.api_v4) return this.api_v4;
+
+    const possible_bases = await Promise.all(
+      pathname
+        .split('/')
+        .filter(Boolean)
+        .map(async (_, index, array) => {
+          const components = [...array.slice(0, index), 'api', 'v4'];
+          const path = this.repo_origin + '/' + components.join('/');
+          try {
+            if ((await this.request({ endpoint: `${path}/version` })).version)
+              return path;
+          } catch (error) {}
+        })
+    );
+
+    const detected_base = possible_bases.find(String);
+    if (detected_base) {
+      this.api_v4 = detected_base;
+      return this.api_v4;
+    }
+
+    throw new Error('GitLab API not found');
   }
 
   async comment_create(opts = {}) {
     const { project_path } = this;
     const { commit_sha, report } = opts;
 
-    const endpoint = `/projects/${project_path}/repository/commits/${commit_sha}/comments`;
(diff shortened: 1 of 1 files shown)


src/drivers/gitlab.js
```diff
@@ -45,7 +72,7 @@ class Gitlab {
 
   async upload(opts = {}) {
     const { project_path, repo } = this;
-    const endpoint = `/projects/${project_path}/uploads`;
+    const endpoint = `${await this.detect_api_v4()}/projects/${project_path}/uploads`;
```
DavidGOrtega, line 75: Why not leave this and subsequent methods intact and recalculate the api_v4 within the request function ?

DavidGOrtega (changes requested):

src/drivers/gitlab.js
0x2b3bfa0: Because we're also using the `request` function to detect the right address. Your suggested modification is rather elegant, but we would need to resolve this circular dependency somehow.

DavidGOrtega: You could pass endpoint as you where or url in opts. If endpoint is passed you call the detect_api... that is returning the proper one if its calculated if url is passed you just take the url

DavidGOrtega: @0x2b3bfa0

0x2b3bfa0: Nice! 🙏

0x2b3bfa0:
@DavidGOrtega, please take a look to the new changes.

src/drivers/gitlab.js
```diff
@@ -149,13 +171,18 @@ class Gitlab {
   }
 
   async request(opts = {}) {
-    const { token, api_v4 } = this;
+    const { token } = this;
     const { endpoint, method = 'GET', body, raw } = opts;
+    let { url } = opts;
 
-    if (!endpoint) throw new Error('Gitlab API endpoint not found');
+    if (endpoint) {
```
DavidGOrtega, line 178:
```js
if (endpoint) {
      this.api_v4 = this.api_v4 || (await this.detect_api_v4());
      url = `${this.api_v4}${endpoint}`;
    } 

if (!url) throw new Error('Gitlab API url not found');
```

DavidGOrtega: dryer and is a validation in any circumstance

DavidGOrtega (changes requested):

src/drivers/gitlab.js
DavidGOrtega: @0x2b3bfa0

0x2b3bfa0: 👌 🙏

0x2b3bfa0:
🔔 @DavidGOrtega

DavidGOrtega:
Looks good. Are you going to do a release of merge straight against master? @0x2b3bfa0

0x2b3bfa0:
@DavidGOrtega, unless you want otherwise, I'll release v0.3.3 for consistency.

DavidGOrtega (approved):

PR #468, opened 2021-04-09
hotfix/Gitlab API detect

DavidGOrtega:
This PR fixes the error of not removing the extra GL url base in the project base.

Right now a url like ```http://localhost:8000/xx/ff/yy```  where ```http://localhost:8000/xx``` is the GL base will generate ```http://localhost:8000/xx/api/v4/xx%2ff%2yy``` instead of ```http://localhost:8000/xx/api/v4/2ff%2yy```


src/drivers/gitlab.js
```diff
@@ -19,39 +19,45 @@ class Gitlab {
...
 
-    const { protocol, host, pathname } = new URL(this.repo);
-    this.project_path = encodeURIComponent(pathname.substring(1));
-    this.repo_origin = `${protocol}//${host}`;
+    return project_path;
   }
 
-  async detect_api_v4(opts = {}) {
-    const { pathname } = new URL(this.repo);
+  async detect_api_v4() {
```
0x2b3bfa0, line 33: After your modifications, `detect_api_v4` should probably be renamed to `detect_base` or something similar.

DavidGOrtega: I thought the same but it uses '/api/v4/version'
Im going to setup this hold on

0x2b3bfa0: Note: I chose `detect_base` instead of `detect_api` because we're actually returning the GitLab server base, and not the API base.

src/drivers/gitlab.js
```diff
@@ -176,8 +186,7 @@ class Gitlab {
     let { url } = opts;
 
     if (endpoint) {
-      this.api_v4 = this.api_v4 || (await this.detect_api_v4());
-      url = `${this.api_v4}${endpoint}`;
+      url = `${await this.detect_api()}/api/${API_VER}${endpoint}`;
```
0x2b3bfa0: Here we're awaiting a function from the templated string itself, but the `comment_create` and `upload` functions above use an intermediate variable. Would it make any sense to standardize on a common approach?

DavidGOrtega: I don't choose one over the other just only if I need more space or not typing.
If the line is short enough I use this over using an extra var.

DavidGOrtega: I mean what prevalences is the line length not the variable assignment

src/drivers/gitlab.js
0x2b3bfa0: In the same line, `repo_base` would also be a bit of a stretch, as the address we return represents the GitLab server base. What about `server_base` instead? By the way: this is just nitpicking, please ignore it if you want to.

src/drivers/gitlab.js
0x2b3bfa0: 👌

src/drivers/gitlab.js
DavidGOrtega: I prefer repo_base to be aligned with repo

0x2b3bfa0: 👌

0x2b3bfa0: Please keep in mind that this address doesn't necessarily involve a `repo` context and, sometimes, we're only being fed with an orgainzation address.

DavidGOrtega: I get you, but its an internal variable and sounded to me nicer to relate it to the source.
If weeks later I read repo_base I will relate it to repo to some extend while server_base I will to figure out what was that

0x2b3bfa0: 👌

src/drivers/gitlab.js
```diff
@@ -19,39 +19,48 @@ class Gitlab {
 
     this.token = token;
     this.repo = repo;
+  }
+
+  async project_path() {
+    const api_path = await this.repo_base();
```
0x2b3bfa0, line 25: The same here: maybe `repo_path` instead of `api_path`?

src/drivers/gitlab.js
```diff
@@ -19,39 +19,48 @@ class Gitlab {
...
+    const project_path = encodeURIComponent(
+      this.repo.replace(api_path, '').substr(1)
```
0x2b3bfa0, line 27: I'd rather explicitly remove the first occurence instead of running `replace`, as it could have unintended consequences in some edge cases.

DavidGOrtega: replace only removes the first occurence unless u use regex /g

DavidGOrtega: @0x2b3bfa0

0x2b3bfa0: Yes, but it doesn't neccessarily have to be the beginning of the string. Maybe it's just paranoid, but we could do something like this (demi-serious pseudocode):

```javascript
if (!this.repo.startsWith(api_path)) throw new Error("unbelievable error");
this.repo.substr(api_path.length + 1)
```

DavidGOrtega: I dont follow you. Replace over substr?

0x2b3bfa0: Well, I actually don't care of which one, but I thought that it would be nice to check first that the string actually starts with the substring we're going to remove.

DavidGOrtega: I dont think that defensive programming is not needed here? If that happens is because the repo is not valid.
If repro_base is working (it is) is always returning a subtr of repo. I cant think right now a case where that happens

0x2b3bfa0 (approved):

src/drivers/gitlab.js
0x2b3bfa0: You're right! 🤦  The input comes from a controlled source, and there isn't any obvious case where this could pose any issue. Sorry for the false positive.

0x2b3bfa0:
```console
$ <(date) node ./bin/cml-send-comment.js \
  --repo=http://···/gitlab/root/test \
  --commit-sha=··· \
  --driver=gitlab \
  --token=···
```

PR #493, opened 2021-04-30
Enforce strict argument parsing

0x2b3bfa0:
Previously, yargs was silently ignoring unknown arguments and producing esoteric bugs on edge cases: `--option ---ARGUMENT---` doesn't produce the same result as `--option=---ARGUMENT---` but would be silently parsed without the strict mode.


casperdcl (approved):
Especialy relevant for `--cloud-ssh-private="---BEGIN RSA PRIVATE KEY..."`

Issue #556, opened 2021-05-21
Investigate Terraform CDK as an alternative to custom templates


0x2b3bfa0:
We should probably use [Terraform CDK](https://www.hashicorp.com/blog/cdk-for-terraform-enabling-python-and-typescript-support) instead of the custom [`src/terraform.js`](https://github.com/iterative/cml/blob/4fd2d8b866aed63dd15a5073f9dfafdb51c13544/src/terraform.js) source file.

See #584 and #564 for more information of the current pitfalls.

DavidGOrtega:
@0x2b3bfa0 The idea would be calling our provider inside the runner instead of generating the terraform file?

0x2b3bfa0:
Yes, this could allow us to use the Terraform provider directly from JavaScript.

DavidGOrtega:
I like it if it does not write a TF file like we do, if not not sure about the effort

0x2b3bfa0:
Not useful.

PR #564, opened 2021-05-26
Mask token in Terraform state output

0x2b3bfa0:
Currently, CI/CD platforms mask it from their logs without requiring us to do anything, but we should mask it on our side anyways.

The proper fix would be honoring the sensitive property of fields by using Terraform CDK as per #556 and outputting only what is needed with [output values](https://www.terraform.io/docs/language/values/outputs.html); e.g. the instance address. See https://github.com/iterative/terraform-provider-iterative/pull/92 and https://github.com/iterative/terraform-provider-iterative/pull/120 for more information.


DavidGOrtega (approved):
👍

casperdcl (approved):

PR #574, opened 2021-06-02
Allow access to secrets for external contributors

0x2b3bfa0:
I forgot to explicitly add access to secrets for external contributors with `pull_request_target` but now it should work as expected. Please keep in mind that approving the test-external environment on malicious pull requests may lead to token exfiltration among other niceties.

### DIFF
--- a/.github/workflows/test-deploy.yml
+++ b/.github/workflows/test-deploy.yml
@@ -7,6 +7,10 @@ on:
   pull_request_target:
 jobs:
   lint:
+    environment:
+      ${{ (github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository) &&
+      'test-external' || 'test-internal' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test-deploy.yml
+++ b/.github/workflows/test-deploy.yml
@@ -4,23 +4,27 @@ on:
     - cron: 0 0 * * *
   release:
     types: [published]
-  pull_request:
+  pull_request_target:
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
       - uses: actions/setup-node@v2
       - run: npm ci
       - run: npm run lint
   test:
     environment:
-      ${{ (github.event_name == 'pull_request' &&
+      ${{ (github.event_name == 'pull_request_target' &&
       github.event.pull_request.head.repo.full_name != github.repository) &&
       'test-external' || 'test-internal' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
       - uses: actions/setup-node@v2
       - uses: actions/setup-python@v2
       - run: pip install tensorboard
@@ -42,6 +46,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
       - uses: actions/setup-node@v2
         with:
           registry-url: https://registry.npmjs.org
@@ -73,6 +79,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
           fetch-depth: 0
       - name: Metadata
         id: metadata
--- a/.github/workflows/test-deploy.yml
+++ b/.github/workflows/test-deploy.yml
@@ -6,11 +6,14 @@ on:
     types: [published]
   pull_request_target:
 jobs:
-  lint:
+  authorize:
     environment:
       ${{ (github.event_name == 'pull_request_target' &&
       github.event.pull_request.head.repo.full_name != github.repository) &&
-      'test-external' || 'test-internal' }}
+      'external' || 'internal' }}
+    runs-on: ubuntu-latest
+  lint:
+    needs: authorize
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -20,10 +23,7 @@ jobs:
       - run: npm ci
       - run: npm run lint
   test:
-    environment:
-      ${{ (github.event_name == 'pull_request_target' &&
-      github.event.pull_request.head.repo.full_name != github.repository) &&
-      'test-external' || 'test-internal' }}
+    needs: authorize
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


0x2b3bfa0:
Untested. Should work as intended, but I'm not completely sure. ⌨️ 🐒

.github/workflows/test-deploy.yml
```diff
@@ -4,23 +4,27 @@ on:
     - cron: 0 0 * * *
   release:
     types: [published]
-  pull_request:
+  pull_request_target:
 jobs:
   lint:
     runs-on: ubuntu-latest
```
0x2b3bfa0, line 10:
```suggestion
  lint:
    environment:
      ${{ (github.event_name == 'pull_request_target' &&
      github.event.pull_request.head.repo.full_name != github.repository) &&
      'test-external' || 'test-internal' }}
    runs-on: ubuntu-latest
```
Running `npm` on untrusted code is also dangerous. Let's protect the lint job with the environment.

.github/workflows/test-deploy.yml
```diff
@@ -4,23 +4,31 @@ on:
...
+  pull_request_target:
 jobs:
+  authorize:
+    environment:
+      ${{ (github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository) &&
+      'external' || 'internal' }}
+    runs-on: ubuntu-latest
   lint:
+    needs: authorize
```
casperdcl, line 16: I don't think we need to worry about linting, right?

0x2b3bfa0: Linting implies running `npm ci` on the source branch, controlled by the contributor, and this could lead to controlled code execution.

.github/workflows/test-deploy.yml
```diff
@@ -4,23 +4,31 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
```
casperdcl, line 21: why is this required?

0x2b3bfa0: Because `pull_request_target` would checkout the target branch by default, not the source branch.

.github/workflows/test-deploy.yml
0x2b3bfa0: https://securitylab.github.com/research/github-actions-preventing-pwn-requests

.github/workflows/test-deploy.yml
casperdcl: don't get it. Are you saying `actions/checkout@v2` needs configuring for `on.*` (including specifically config for `ref` for `on.pull_request_target`)? I doubt it.

casperdcl: After some reading it looks like `actions/checkout` *is* indeed broken. Need to do some more reading before I'd be happy to accept this. Grrr, GitHub.

0x2b3bfa0: Not actually broken: https://securitylab.github.com/research/github-actions-preventing-pwn-requests

0x2b3bfa0: It just uses the target branch instead of the source branch to avoid running privileged code from untrusted sources.

casperdcl: > Not actually broken: [securitylab.github.com/research/github-actions-preventing-pwn-requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests)

that link is amongst many things the precise definition of broken :)

0x2b3bfa0: > _that link is amongst many things the precise definition of broken_

If you mean literally broken... it works on my computer. ™ 🤔  Please try [reading it from Archive.org]( https://web.archive.org/web/20210426102036/https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

.github/workflows/test-deploy.yml
```diff
@@ -4,23 +4,31 @@ on:
     - cron: 0 0 * * *
   release:
     types: [published]
-  pull_request:
+  pull_request_target:
```
casperdcl, line 7:
```suggestion
  pull_request_target:
defaults:
  environment:
    ${{ (github.event_name == 'pull_request_target' &&
    github.event.pull_request.head.repo.full_name != github.repository) &&
    'external' || 'internal' }}
```

.github/workflows/test-deploy.yml
```diff
@@ -4,23 +4,31 @@ on:
...
 jobs:
+  authorize:
+    environment:
+      ${{ (github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository) &&
+      'external' || 'internal' }}
+    runs-on: ubuntu-latest
```
casperdcl, line 14:
```suggestion
```

.github/workflows/test-deploy.yml
casperdcl, line 16:
```suggestion
```

.github/workflows/test-deploy.yml
```diff
@@ -4,23 +4,31 @@ on:
...
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
       - uses: actions/setup-node@v2
       - run: npm ci
       - run: npm run lint
   test:
-    environment:
-      ${{ (github.event_name == 'pull_request' &&
-      github.event.pull_request.head.repo.full_name != github.repository) &&
-      'test-external' || 'test-internal' }}
+    needs: authorize
```
casperdcl, line 26:
```suggestion
```

casperdcl (changes requested):
does `defaults.environment` work?

0x2b3bfa0:
> does `defaults.environment` work?

No, it [doesn't](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaults) and, in case it did, it would require approval for each individaul step.

casperdcl (approved):
GH business logic of `actions/checkout` is broken due to their (ab)use of `on.pull_request_target` for two different purposes, one of which requires checkout `with.ref: pull_request.head.sha`. No clean work-around until they implement something like `on.pull_request_target_v2`

:cry:

casperdcl:
worth writing a short blog post about @0x2b3bfa0? (/CC @shcheklein)

0x2b3bfa0:
> _worth writing a short blog post about @0x2b3bfa0?_

(I hope you didn't omit a comma before “@0x2b3bfa0” on purpose) 🥶

0x2b3bfa0:
While setting [“require approval for all the outside collaborators”](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-required-approval-for-workflows-from-public-forks) would be a cleaner solution, it doesn’t have any effect for the `pull_request_target` event:

> [_Workflows triggered by `pull_request_target` events are run in the context of the base branch. Since the base branch is considered trusted, workflows triggered by these events will always run, regardless of approval settings._](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

[Environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) are needed to require [manual approval](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) for external contributions. Otherwise, contributors could exfiltrate the repository secrets.

### See also
* https://securitylab.github.com/research/github-actions-preventing-pwn-requests
* https://github.com/iterative/mlem/pull/318

casperdcl:
> hope you didn't omit a comma

xkcd#191

fxmarty:
Hi @0x2b3bfa0, this approach seem to work decently well. Isn't it still prone to a race condition in which the attacker may push new changes after the workflow was approved? As explained in: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Could the PR be updated with malicious code just after the approval?

fxmarty:
Thank you @0x2b3bfa0! It works well indeed!

fxmarty:
The only issue I have now is that I have external collaborators with write access to my repo. I would like them to be able to run automatically the workflow - but it appears only member from the `external` environment can (and there can only be 6 members/teams there).

0x2b3bfa0:
Maybe something like this?

````
    environment:
      ${{ github.event_name == 'pull_request_target' 
       && github.event.pull_request.head.repo.full_name != github.repository
       && github.event.pull_request.user.login != 'authorized_user_1'
       && github.event.pull_request.user.login != 'authorized_user_2'
       && github.event.pull_request.user.login != 'authorized_user_n'
       && 'external'
       || 'internal'
       }}
```

Beware of user name changes, though!

Issue #538, opened 2021-05-19
fix/upgrade integration tests


DavidGOrtega:
- tests sometimes pass despite introducing breaking changes
  + The main reason was having introduced a gyp lib that needed sudo permissions breaking the cml service inside the CML aws image managed by our terraform provider
- should use external repo & `send-check` to
  + avoid spamming CML with comments
  + avoid too many token permissions

We need:
 - Integration tests that take the whole picture 
    - testbot should do the trick however the CML image is picking from master

0x2b3bfa0:
> _however the CML image is picking from master <sup>[citation required]</sup>_

As far as I know, we don't do that anymore <sup>[[1](https://github.com/iterative/terraform-provider-iterative/blob/5f62a02d428eb9f6961e1f0721ec76364d7fe347/iterative/resource_runner.go#L275), [2](https://github.com/iterative/cml/blob/d716d3072d4d9829865d95f67bdf79cc7c9c5a62/Dockerfile#L78-L79)]</sup>

0x2b3bfa0:
> _testbot should do the trick however the CML image is picking from master_

Do you mean that the [checkbot](https://github.com/iterative/cml/blob/master/.github/workflows/checkbot.yml) test image is being built from the latest changes in the branch that triggered the workflow?

DavidGOrtega:
No, what I say is that despite that we are testing everything when we launch a cloud runner it is going to use "live" cml not "dev", hence we are not going to notice those issues and testing with cloud runners is unsecure

0x2b3bfa0:
> [_As per #527, I would tend to avoid integration tests whenever possible and detect these issues by resorting to simpler unit tests._](https://github.com/iterative/cml/issues/550#issuecomment-845166152)

0x2b3bfa0:
### Suggested action points

- [ ] Use [custom setup and teardown](https://jestjs.io/docs/setup-teardown#one-time-setup) functions to create and destroy ephemeral test repositories on GitHub, GitLab and Bitbucket where we can run each test in a clean slate and without getting extra notifications. Most of these services can be executed inside GitHub Actions as [service containers](https://docs.github.com/es/actions/guides/about-service-containers#about-service-containers); see the [`gitlab.yml`](https://github.com/iterative/cml/blob/899c6d2e3eaffc66d12492deffa658e8430f1c66/.github/workflows/gitlab.yml
[...]

PR #581, opened 2021-06-08
Add `cml-send-comment --update`

0x2b3bfa0:
Running `cml-send-comment` with the newly introduced `--update` flag will update the last CML comment instead of creating a new one. Comment authorship is determined by checking an invisible _watermark_ [sic] added by CML to every machine-made comment. Watermarks are just void images appended with Markdown at the end of comments.

### Supported platforms

- [X] GitHub
- [ ] GitLab<sup>1</sup>
- [X] Bitbucket<sup>2</sup>

***

<sup>1</sup> Blocked by [gitlab-org/gitlab#333000](https://gitlab.com/gitlab-org/gitlab/-/issues/333000); there is no known workaround.
<sup>2</sup> Requires user interaction for the first time, as per https://github.com/iterative/cml/pull/581#issuecomment-861898718.

### DIFF
--- a/bin/cml-send-comment.js
+++ b/bin/cml-send-comment.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+const print = console.log;
 console.log = console.error;
 
 const fs = require('fs').promises;
@@ -11,7 +12,7 @@ const run = async (opts) => {
   const path = opts._[0];
   const report = await fs.readFile(path, 'utf-8');
   const cml = new CML(opts);
-  console.log(await cml.commentCreate({ ...opts, report }));
+  print(await cml.commentCreate({ ...opts, report }));
 };
 
 const opts = yargs
--- a/bin/cml-send-comment.js
+++ b/bin/cml-send-comment.js
@@ -44,7 +44,7 @@ const opts = yargs
     'Personal access token to be used. If not specified is extracted from ENV REPO_TOKEN.'
   )
   .default('driver')
-  .choices('driver', ['github', 'gitlab'])
+  .choices('driver', ['github', 'gitlab', 'bitbucket'])
   .describe('driver', 'If not specify it infers it from the ENV.')
   .help('h')
   .demand(1).argv;
--- a/bin/cml-send-comment.js
+++ b/bin/cml-send-comment.js
@@ -36,7 +36,7 @@ const opts = yargs
   .default('token')
   .describe(
     'token',
-    'Personal access token to be used. If not specified in extracted from ENV REPO_TOKEN.'
+    'Personal access token to be used. If not specified is extracted from ENV REPO_TOKEN.'
   )
   .default('driver')
   .choices('driver', ['github', 'gitlab'])
--- a/bin/cml-send-comment.js
+++ b/bin/cml-send-comment.js
@@ -11,7 +11,7 @@ const run = async (opts) => {
   const path = opts._[0];
   const report = await fs.readFile(path, 'utf-8');
   const cml = new CML(opts);
-  await cml.commentCreate({ ...opts, report });
+  console.log(await cml.commentCreate({ ...opts, report }));
 };
 
 const opts = yargs
@@ -23,6 +23,11 @@ const opts = yargs
     'Commit SHA linked to this comment. Defaults to HEAD.'
   )
   .alias('commit-sha', 'head-sha')
+  .boolean('update')
+  .describe(
+    'update',
+    'Update the last CML comment (if any) instead of creating a new one'
+  )
   .boolean('rm-watermark')
   .describe(
     'rm-watermark',
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -31,7 +31,8 @@ class BitBucketCloud {
       prs = await this.paginatedRequest({ endpoint: getPrEndpoint });
     } catch (err) {
       if (err.message === 'Not Found Resource not found')
-        err.message = 'Click \'Go to pull request\' on any commit details page to enable this API'
+        err.message =
+          "Click 'Go to pull request' on any commit details page to enable this API";
       throw err;
     }
 
@@ -61,7 +62,7 @@ class BitBucketCloud {
       }
     }
 
-        const commitEndpoint = `/repositories/${projectPath}/commit/${commitSha}/comments/`;
+    const commitEndpoint = `/repositories/${projectPath}/commit/${commitSha}/comments/`;
 
     const existingCommmit = (
       await this.paginatedRequest({ endpoint: commitEndpoint, method: 'GET' })
@@ -81,7 +82,7 @@ class BitBucketCloud {
         method: update && existingCommmit ? 'PUT' : 'POST',
         body: JSON.stringify({ content: { raw: report } })
       })
-    ).links.html.href;    
+    ).links.html.href;
   }
 
   async checkCreate() {
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -52,7 +52,11 @@ class BitBucketCloud {
       const getPrEndpoint = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
       prs = await this.paginatedRequest({ endpoint: getPrEndpoint });
     } catch (err) {
-      if (err.message !== 'Not Found Resource not found') throw err;
+      if (err.message === 'Not Found Resource not found')
+        console.warn(
+          "Can't create a pull request comment: the Pull Request Commit Links application has not been installed."
+        );
+      else throw err;
     }
 
     if (prs && prs.length) {
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -201,18 +201,18 @@ class BitBucketCloud {
 
   async paginatedRequest(opts = {}) {
     const { method = 'GET', body } = opts;
-    const result = await this.request(opts);
+    const {next, values} = await this.request(opts);
 
-    if (result.next) {
-      const next = await this.paginatedRequest({
-        url: result.next,
+    if (next) {
+      const nextValues = await this.paginatedRequest({
+        url: next,
         method,
         body
       });
-      result.values.push(...next);
+      values.push(...nextValues);
     }
 
-    return result.values;
+    return values;
   }
 
   get sha() {
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -40,7 +40,7 @@ class BitBucketCloud {
       for (const pr of prs) {
         // Append a watermark to the report with a link to the commit
         const commitLink = commitSha.substr(0, 7);
-        const longReport = `${commitLink}   \n${report}`;
+        const longReport = `${commitLink}\n\n${report}`;
         const prBody = JSON.stringify({ content: { raw: longReport } });
 
         // Write a comment on the PR
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -184,10 +184,11 @@ class BitBucketCloud {
       'Content-Type': 'application/json',
       Authorization: 'Basic ' + `${token}`
     };
-    const response = await fetch(
-      url || `${api}${endpoint}`,
-      { method, headers, body }
-    );
+    const response = await fetch(url || `${api}${endpoint}`, {
+      method,
+      headers,
+      body
+    });
 
     if (response.status > 300) {
       const {
@@ -201,7 +202,7 @@ class BitBucketCloud {
 
   async paginatedRequest(opts = {}) {
     const { method = 'GET', body } = opts;
-    const {next, values} = await this.request(opts);
+    const { next, values } = await this.request(opts);
 
     if (next) {
       const nextValues = await this.paginatedRequest({
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -24,28 +24,6 @@ class BitBucketCloud {
     const { projectPath } = this;
     const { commitSha, report, update, watermark } = opts;
 
-    const commitEndpoint = `/repositories/${projectPath}/commit/${commitSha}/comments/`;
-
-    const existingCommmit = (
-      await this.paginatedRequest({ endpoint: commitEndpoint, method: 'GET' })
-    )
-      .filter((comment) => {
-        const { content: { raw = '' } = {} } = comment;
-        return raw.endsWith(watermark);
-      })
-      .sort((first, second) => first.id < second.id)
-      .pop();
-
-    const commitOutput = (
-      await this.request({
-        endpoint:
-          commitEndpoint +
-          (update && existingCommmit ? existingCommmit.id : ''),
-        method: update && existingCommmit ? 'PUT' : 'POST',
-        body: JSON.stringify({ content: { raw: report } })
-      })
-    ).links.html.href;
-
     // Check for a corresponding PR. If it exists, also put the comment there.
     let prs;
     try {
@@ -53,10 +31,8 @@ class BitBucketCloud {
       prs = await this.paginatedRequest({ endpoint: getPrEndpoint });
     } catch (err) {
       if (err.message === 'Not Found Resource not found')
-        console.warn(
-          "Can't create a pull request comment: the Pull Request Commit Links application has not been installed."
-        );
-      else throw err;
+        err.message = 'Click \'Go to pull request\' on any commit details page to enable this API'
+      throw err;
     }
 
     if (prs && prs.length) {
@@ -85,7 +61,27 @@ class BitBucketCloud {
       }
     }
 
-    return commitOutput;
+        const commitEndpoint = `/repositories/${projectPath}/commit/${commitSha}/comments/`;
+
+    const existingCommmit = (
+      await this.paginatedRequest({ endpoint: commitEndpoint, method: 'GET' })
+    )
+      .filter((comment) => {
+        const { content: { raw = '' } = {} } = comment;
+        return raw.endsWith(watermark);
+      })
+      .sort((first, second) => first.id < second.id)
+      .pop();
+
+    return (
+      await this.request({
+        endpoint:
+          commitEndpoint +
+          (update && existingCommmit ? existingCommmit.id : ''),
+        method: update && existingCommmit ? 'PUT' : 'POST',
+        body: JSON.stringify({ content: { raw: report } })
+      })
+    ).links.html.href;    
   }
 
   async checkCreate() {
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -22,39 +22,62 @@ class BitBucketCloud {
 
   async commentCreate(opts = {}) {
     const { projectPath } = this;
-    const { commitSha, report } = opts;
+    const { commitSha, report, update, watermark } = opts;
 
-    // Make a comment in the commit
     const commitEndpoint = `/repositories/${projectPath}/commit/${commitSha}/comments/`;
-    const commitBody = JSON.stringify({ content: { raw: report } });
-    const commitOutput = await this.request({
-      endpoint: commitEndpoint,
-      method: 'POST',
-      body: commitBody
-    });
+
+    const existingCommmit = (
+      await this.paginatedRequest({ endpoint: commitEndpoint, method: 'GET' })
+    )
+      .filter(
+        (comment) =>
+          comment.content.raw && comment.content.raw.endsWith(watermark)
+      )
+      .sort((first, second) => first.id < second.id)
+      .pop();
+
+    const commitOutput = (
+      await this.request({
+        endpoint:
+          commitEndpoint +
+          (update && existingCommmit ? existingCommmit.id : ''),
+        method: update && existingCommmit ? 'PUT' : 'POST',
+        body: JSON.stringify({ content: { raw: report } })
+      })
+    ).links.html.href;
 
     // Check for a corresponding PR. If it exists, also put the comment there.
-    const getPrEndpt = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
-    const { values: prs } = await this.request({ endpoint: getPrEndpt });
+    let prs;
+    try {
+      const getPrEndpoint = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
+      prs = await this.paginatedRequest({ endpoint: getPrEndpoint });
+    } catch (err) {
+      if (err.message !== 'Not Found Resource not found') throw err;
+    }
 
     if (prs && prs.length) {
       for (const pr of prs) {
-        try {
-          // Append a watermark to the report with a link to the commit
-          const commitLink = commitSha.substr(0, 7);
-          const longReport = `${commitLink}   \n${report}`;
-          const prBody = JSON.stringify({ content: { raw: longReport } });
+        // Append a watermark to the report with a link to the commit
+        const commitLink = commitSha.substr(0, 7);
+        const longReport = `${commitLink}   \n${report}`;
+        const prBody = JSON.stringify({ content: { raw: longReport } });
 
-          // Write a comment on the PR
-          const prEndpoint = `/repositories/${projectPath}/pullrequests/${pr.id}/comments`;
-          await this.request({
-            endpoint: prEndpoint,
-            method: 'POST',
-            body: prBody
-          });
-        } catch (err) {
-          console.debug(err.message);
-        }
+        // Write a comment on the PR
+        const prEndpoint = `/repositories/${projectPath}/pullrequests/${pr.id}/comments/`;
+        const existingPr = (
+          await this.paginatedRequest({ endpoint: prEndpoint, method: 'GET' })
+        )
+          .filter(
+            (comment) =>
+              comment.content.raw && comment.content.raw.endsWith(watermark)
+          )
+          .sort((first, second) => first.id < second.id)
+          .pop();
+        await this.request({
+          endpoint: prEndpoint + (update && existingPr ? existingPr.id : ''),
+          method: update && existingPr ? 'PUT' : 'POST',
+          body: prBody
+        });
       }
     }
 
@@ -150,14 +173,14 @@ class BitBucketCloud {
 
   async request(opts = {}) {
     const { token, api } = this;
-    const { endpoint, method = 'GET', body } = opts;
-
-    if (!endpoint) throw new Error('BitBucket Cloud API endpoint not found');
+    const { fullEndpoint, endpoint, method = 'GET', body } = opts;
+    if (!(endpoint || fullEndpoint))
+      throw new Error('BitBucket Cloud API endpoint not found');
     const headers = {
       'Content-Type': 'application/json',
       Authorization: 'Basic ' + `${token}`
     };
-    const url = `${api}${endpoint}`;
+    const url = fullEndpoint || `${api}${endpoint}`;
     const response = await fetch(url, { method, headers, body });
 
     if (response.status > 300) {
@@ -168,6 +191,22 @@ class BitBucketCloud {
     }
 
     return await response.json();
+  }
+
+  async paginatedRequest(opts = {}) {
+    const { method = 'GET', body } = opts;
+    const result = await this.request(opts);
+
+    if (result.next) {
+      const next = await this.paginatedRequest({
+        fullEndpoint: result.next,
+        method,
+        body
+      });
+      result.values.push(...next);
+    }
+
+    return result.values;
   }
 
   get sha() {
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -177,15 +177,17 @@ class BitBucketCloud {
 
   async request(opts = {}) {
     const { token, api } = this;
-    const { fullEndpoint, endpoint, method = 'GET', body } = opts;
-    if (!(endpoint || fullEndpoint))
+    const { url, endpoint, method = 'GET', body } = opts;
+    if (!(url || endpoint))
       throw new Error('BitBucket Cloud API endpoint not found');
     const headers = {
       'Content-Type': 'application/json',
       Authorization: 'Basic ' + `${token}`
     };
-    const url = fullEndpoint || `${api}${endpoint}`;
-    const response = await fetch(url, { method, headers, body });
+    const response = await fetch(
+      url || `${api}${endpoint}`,
+      { method, headers, body }
+    );
 
     if (response.status > 300) {
       const {
@@ -203,7 +205,7 @@ class BitBucketCloud {
 
     if (result.next) {
       const next = await this.paginatedRequest({
-        fullEndpoint: result.next,
+        url: result.next,
         method,
         body
       });
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -29,10 +29,10 @@ class BitBucketCloud {
     const existingCommmit = (
       await this.paginatedRequest({ endpoint: commitEndpoint, method: 'GET' })
     )
-      .filter(
-        (comment) =>
-          comment.content.raw && comment.content.raw.endsWith(watermark)
-      )
+      .filter((comment) => {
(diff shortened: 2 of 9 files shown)


casperdcl:
Given https://github.com/iterative/cml/wiki/Backend-Supported-Features we could still merge anyway

0x2b3bfa0:
> [_Pull Request Commit Links app must be installed first before using this API; installation automatically occurs when 'Go to pull request' is clicked from the web interface for a commit's details._](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/commit/%7Bcommit%7D/pullrequests) (Clinical definition of insanity)

bin/cml-send-comment.js
```diff
@@ -11,7 +11,7 @@ const run = async (opts) => {
   const path = opts._[0];
   const report = await fs.readFile(path, 'utf-8');
   const cml = new CML(opts);
-  await cml.commentCreate({ ...opts, report });
+  console.log(await cml.commentCreate({ ...opts, report }));
```
0x2b3bfa0, line 14: Now the `cml-send-comment` command will return a link to the comment for consistency with tools like `cml-pr` or `gh` that already do that. It seems like a good addition to me, but is a bit out of scope for this pull request.

bin/cml-send-comment.js
```diff
@@ -36,10 +41,10 @@ const opts = yargs
   .default('token')
   .describe(
     'token',
-    'Personal access token to be used. If not specified in extracted from ENV REPO_TOKEN.'
+    'Personal access token to be used. If not specified is extracted from ENV REPO_TOKEN.'
   )
   .default('driver')
-  .choices('driver', ['github', 'gitlab'])
+  .choices('driver', ['github', 'gitlab', 'bitbucket'])
```
0x2b3bfa0, line 47: Bitbucket support wasn't enabled at the command level. Was it intentional? 🤔

src/drivers/github.test.js
```diff
@@ -1,4 +1,4 @@
-jest.setTimeout(20000);
+jest.setTimeout(40000);
```
0x2b3bfa0, line 1: Had to increase all these timeouts for tests to run correctly.

src/drivers/github.js
```diff
@@ -70,18 +70,37 @@ class Github {
...
+      ).data.html_url;
+    } else {
+      return (
+        await rest.repos.createCommitComment({
+          ...ownerRepo({ uri: this.repo }),
+          commit_sha: commitSha,
+          body
+        })
+      ).data.html_url;
+    }
```
0x2b3bfa0, line 103: @DavidGOrtega dislikes this kind of structure with several `return` statements that run exclusively from the different branches of a conditional. I find that, in this specific case, the intention is clearer than with the terse alternative, but I don't care that much if any of you prefers to write it differently.

casperdcl:
> > [_Pull Request Commit Links app must be installed first before using this API; installation automatically occurs when 'Go to pull request' is clicked from the web interface for a commit's details._](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/commit/%7Bcommit%7D/pullrequests) (Clinical definition of insanity)

can we

- auto-detect this and emit a warning
- document this (probably in https://github.com/iterative/cml/wiki/cml-send-comment)

src/drivers/bitbucket_cloud.js
```diff
@@ -22,39 +22,62 @@ class BitBucketCloud {
...
 
     // Check for a corresponding PR. If it exists, also put the comment there.
-    const getPrEndpt = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
-    const { values: prs } = await this.request({ endpoint: getPrEndpt });
+    let prs;
+    try {
+      const getPrEndpoint = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
+      prs = await this.paginatedRequest({ endpoint: getPrEndpoint });
+    } catch (err) {
+      if (err.message !== 'Not Found Resource not found') throw err;
```
0x2b3bfa0, line 55:
```suggestion
      if (err.message === 'Not Found Resource not found')
        console.warn(
          "Can't create a pull request comment: the Pull Request Commit Links application has not been installed."
        );
      else throw err;
```

0x2b3bfa0:
Done! The documentation part is a bit terse, but probably good enough for now. Feel free to expand it if you want to.

src/drivers/github.js
DavidGOrtega: I would say its never clearer and adds an indentation, if its the last thing to be done

bin/cml-send-comment.js
DavidGOrtega, line 14: should not have a print statement? Why are we printing in stderror?

src/drivers/github.js
0x2b3bfa0:
```suggestion
    if (update && existing)
      return (
        await rest.repos.updateCommitComment({
          ...ownerRepo({ uri: this.repo }),
          comment_id: existing.id,
          body
        })
      ).data.html_url;
      
    return (
      await rest.repos.createCommitComment({
        ...ownerRepo({ uri: this.repo }),
        commit_sha: commitSha,
        body
      })
    ).data.html_url;
```

Would you prefer it like that?

bin/cml-send-comment.js
DavidGOrtega: thats very weird! Indeed bitbucket should not work and it does!
strict seems not be be applying here?

src/drivers/bitbucket_cloud.js
```diff
@@ -22,39 +22,66 @@ class BitBucketCloud {
...
+        const prBody = JSON.stringify({ content: { raw: longReport } });
+
+        // Write a comment on the PR
+        const prEndpoint = `/repositories/${projectPath}/pullrequests/${pr.id}/comments/`;
+        const existingPr = (
+          await this.paginatedRequest({ endpoint: prEndpoint, method: 'GET' })
+        )
+          .filter(
+            (comment) =>
+              comment.content.raw && comment.content.raw.endsWith(watermark)
```
DavidGOrtega, line 76: can we deestructure here?
```js
const { content: { raw = '' } = {} } = comment;
raw && raw.endsWith(watermark)
```

src/drivers/bitbucket_cloud.js
```diff
@@ -22,39 +22,66 @@ class BitBucketCloud {
...
+          )
+          .sort((first, second) => first.id < second.id)
+          .pop();
+        await this.request({
+          endpoint: prEndpoint + (update && existingPr ? existingPr.id : ''),
```
DavidGOrtega, line 81: can we assing first the vars and then:

```js
await this.request({ endpoint, method, body })
```

src/drivers/github.js
0x2b3bfa0: Some extensive discussions on both appreaches:

https://stackoverflow.com/questions/9191388/it-is-more-efficient-to-use-if-return-return-or-if-else-return
https://softwareengineering.stackexchange.com/questions/127128/what-is-the-best-style-for-functions-with-multiple-returns-and-if-else-statement
https://softwareengineering.stackexchange.com/questions/157407/best-practice-on-if-return

It looks like this kind of choice is the ideal breeding ground for a holy war with a copious arsenal for both bands. 😅  As stated, I don't have any strong opinion on this and I'm happy to adopt the alternate style for consistency with the project. Feel free to apply the suggestion.

src/drivers/bitbucket_cloud.js
```diff
@@ -150,14 +177,14 @@ class BitBucketCloud {
 
   async request(opts = {}) {
     const { token, api } = this;
-    const { endpoint, method = 'GET', body } = opts;
-
-    if (!endpoint) throw new Error('BitBucket Cloud API endpoint not found');
+    const { fullEndpoint, endpoint, method = 'GET', body } = opts;
+    if (!(endpoint || fullEndpoint))
```
DavidGOrtega, line 181: would not be better uri or url vs fullEndpoint?

src/drivers/bitbucket_cloud.js
```diff
@@ -170,6 +197,22 @@ class BitBucketCloud {
     return await response.json();
   }
 
+  async paginatedRequest(opts = {}) {
+    const { method = 'GET', body } = opts;
+    const result = await this.request(opts);
+
+    if (result.next) {
+      const next = await this.paginatedRequest({
+        fullEndpoint: result.next,
```
DavidGOrtega, line 206: assign previously

src/drivers/bitbucket_cloud.js
```diff
@@ -170,6 +197,22 @@ class BitBucketCloud {
     return await response.json();
   }
 
+  async paginatedRequest(opts = {}) {
+    const { method = 'GET', body } = opts;
+    const result = await this.request(opts);
```
DavidGOrtega, line 202: desstructure

```js
const { next: fullEndpoint } =  await this.request(opts);
```

src/drivers/bitbucket_cloud.js
```diff
@@ -170,6 +197,22 @@ class BitBucketCloud {
...
+
+    if (result.next) {
+      const next = await this.paginatedRequest({
+        fullEndpoint: result.next,
+        method,
+        body
+      });
+      result.values.push(...next);
```
DavidGOrtega, line 210: { values: } ...

src/drivers/github.js
```diff
@@ -70,18 +70,37 @@ class Github {
...
-    ).repos.createCommitComment({
-      ...ownerRepo({ uri: this.repo }),
-      body,
-      commit_sha: commitSha
-    });
-
-    return commitUrl;
+    const { report: body, commitSha, update, watermark } = opts;
+
+    const rest = octokit(this.token, this.repo);
```
DavidGOrtega, line 75: deestructure
```js
const {paginate, repos } = ...
```

DavidGOrtega (changes requested):

bin/cml-send-comment.js
DavidGOrtega: Its printing in stderror. It should have 
const print = console.log

bin/cml-send-comment.js
0x2b3bfa0: > _Indeed Bitbucket should not work and it does!_

It does work, at least for `cml-send-comment` 🙃  I thoroughly tested it yesterday and it also passes the automated checks.

> _strict seems not be be applying here?_

What do you mean?

src/drivers/github.js
DavidGOrtega: In my university this was not a holy war at all.
If you did so or did bolean === 'true' you where automatically wasted 😁

bin/cml-send-comment.js
0x2b3bfa0: Nice catch!

src/drivers/bitbucket_cloud.js
0x2b3bfa0: Yes, we can, but I don't see any advantage over the current approach, which, to my mind, seems a lot clearer. Feel free to unresolve if you think otherwise.

DavidGOrtega: Nope. Its js style and we have to stick with a style. Aside that is far larger.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: Yes

src/drivers/bitbucket_cloud.js
0x2b3bfa0: As you mentioned [here](https://iterativeai.slack.com/archives/C01900GSB4J/p1623844680401200), we're using [this convention](https://github.com/airbnb/javascript#destructuring).  Nevertheless, I'm failing to see how the proposed solution could be shorter or more readable than the original.

```suggestion
              {
                const { content: { raw = '' } = {} } = comment;
                return raw.endsWith(watermark);
              }
```

Feel free to update the code under your own responsibility. ☠️  I still think that [this](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds) prevails over the style guide. 😉

src/drivers/bitbucket_cloud.js
0x2b3bfa0: We _can_, but having a new variable named `endpoint` in a context where we are going to need both might be counterintuitive. Unless you have a strong preference for the suggested approach, I would prefer to keep it as is.

src/drivers/bitbucket_cloud.js
DavidGOrtega: > Nevertheless, I'm failing to see how the proposed solution could be shorter or more readable than the original.

shorter the line @0x2b3bfa0 and dryer

```js
myobject.oneproperty.amazing && myobject.oneproperty.amazing.istrulyamazing && myobject.oneproperty.amazing.istrulyamazing()
```

its way larger than having the solution destructured. This is something that tends to happen in Object based languages and JS is an Object based language (where even a function is an Object). To avoid this ES6 included deestructuring  as a first citizen. So even in a case where you don't see the benefit of it, being consistent allows you to write it without think on it. 

P.D. Im  very sorry to have a little mind

src/drivers/bitbucket_cloud.js
DavidGOrtega: same as before, is a matter of style

src/drivers/bitbucket_cloud.js
DavidGOrtega: why not?

src/drivers/bitbucket_cloud.js
DavidGOrtega: Which one is good? up or down?

src/drivers/bitbucket_cloud.js
0x2b3bfa0: > _and dryer_

Ouch! 😄  That's unquestionably true.

> _its way larger than having the solution destructured. This is something that tends to happen in Object based languages and JS is an Object based language (where even a function is an Object). To avoid this ES6 included deestructuring as a first citizen. So even in a case where you don't see the benefit of it, being consistent allows you to write it without think on it._

Once and if we move to Node JS 14+ we can use `comment?.content?.raw?.endsWith(watermark)` thanks to [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining); it's the really idiomatic solution to this issue.

Anyhow, I don't want to spend your time in sterile discussions. 🤷🏼  If you think it's the way to go, I'm fine with it.

> _P.D. Im very sorry to have a little mind_

Don't take it personal, it's about approaches, not about minds.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: 🤷🏼  As usual, feel free to update it to fit the style.

src/drivers/bitbucket_cloud.js
```diff
@@ -199,18 +202,18 @@ class BitBucketCloud {
...
-      const next = await this.paginatedRequest({
-        fullEndpoint: result.next,
+    if (next) {
+      const nextValues = await this.paginatedRequest({
+        url: next,
         method,
         body
       });
-      result.values.push(...next);
+      values.push(...nextValues);
```
DavidGOrtega, line 213: why the three dots?

```js
values.push({ ...nextValues });
```

or (if you dont need a deep copy)

values.push(nextValues);

does this even work 🤔 ?

src/drivers/bitbucket_cloud.js
0x2b3bfa0: If you mean my double (uo|down)vote, both are intended. 😄  I wanted to say that I liked the idea of destructuring there, but that the suggested snippet was wrong. Nevermind.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: Your suggested `values.push(nextValues);` doesn't produce the same effect: we intend to concatenate arrays, not nest them. And `values.push({ ...nextValues });` doesn't behave as expected with arrays either.

src/drivers/bitbucket_cloud.js
DavidGOrtega: > Anyhow, I don't want to spend your time in sterile discussions. 🤷🏼 If you think it's the way to go, I'm fine with it.

It's good to spend time on this. To be working on the same page... its never about me or my personal preferences. If we don't like the style lets chose one and stick with it.

src/drivers/bitbucket_cloud.js
DavidGOrtega: ah! I thought it was an object.

bin/cml-send-comment.js
DavidGOrtega: yargs has strict, if we were setting a weird value it should not like it. Thats the goal of choices here

src/drivers/bitbucket_cloud.js
0x2b3bfa0: > _It's good to spend time on this. To be working on the same page... its never about me or my personal preferences. If we don't like the style lets chose one and stick with it._

Awesome! I don't want to impose my personal preferences either, as it would be as unfair and impractical; let's ping @casperdcl so we can take a more plural decision. Destructuring doesn't look that bad either, but it seemed a bit more Perl-ified than the original code. 

In this particular case, I would be really happy to use Node JS 14+ [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead of nested destructurings or infinite checking mantras. Nevertheless, we're still stuck on Node JS 12 and the immediate solution would require us to resort to any of the discussed approaches.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: No, it's the `values` array, that contains many objects. We're using tail recursion to append each `next` page values to the end of the results returned by the previous page.

src/drivers/bitbucket_cloud.js
DavidGOrtega: we have #596 for this

0x2b3bfa0: And https://github.com/iterative/cml/issues/597

src/drivers/bitbucket_cloud.js
0x2b3bfa0: Because I think that checking for `next` conveys better the purpose of the algorithm than checking for `url` in the conditional.

src/drivers/bitbucket_cloud.js
DavidGOrtega: Why was wrong ? I did not follow the code but I saw this

```js
fullEndpoint: result.next,
```

and I inferred this

```js
const { next: fullEndpoint } =  await this.request(opts);
```

where you declare next as fullEndpoint

0x2b3bfa0: It's a tricky one: we also use `result.values` at the end, not only `result.next` as you inferred.

```javascript
const { next, values } =  await this.request(opts);
```

> _This said again without fully reading the code :)_

If you achieve (nearly) linter-level precision without reading the code, I can't imagine what you would do when fully reading the code. 😂

bin/cml-send-comment.js
0x2b3bfa0: And it works as intended. In fact, tests don't invoke the command-lin tool but the `cml` module. That's why it doesn't trip yargs when running Bitbucket tests.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: Closed with https://github.com/iterative/cml/issues/596#issuecomment-862491282 and https://github.com/iterative/cml/pull/581/commits/68cb5a27dbfccea85d9aa1dbe06287cea519c20c

0x2b3bfa0:
Hopefully I won't have to think about John Woods' quote for a long, long time.

bin/cml-send-comment.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+const print = console.log;
```
casperdcl, line 3: seems odd to alias if it's only used once in the whole file

bin/cml-send-comment.js
```diff
@@ -1,5 +1,6 @@
...
 console.log = console.error;
```
casperdcl, line 4: really? why?

src/drivers/bitbucket_cloud.js
```diff
@@ -22,39 +22,66 @@ class BitBucketCloud {
...
-            endpoint: prEndpoint,
-            method: 'POST',
-            body: prBody
-          });
-        } catch (err) {
-          console.debug(err.message);
-        }
+        // Append a watermark to the report with a link to the commit
+        const commitLink = commitSha.substr(0, 7);
+        const longReport = `${commitLink}   \n${report}`;
```
casperdcl, line 66: why the trailing WS?

```suggestion
        const longReport = `${commitLink}\n${report}`;
```

src/drivers/github.test.js
casperdcl: odd. CML+=40, BB+=100, GH+=20. No clear pattern.

bin/cml-send-comment.js
0x2b3bfa0: You're 27 days late: https://github.com/iterative/cml/issues/22#issuecomment-846022553 😈  We use this on all the commands and changing it would be out of scope for this pull request, but it would be nice to choose a less "interesting" approach.

bin/cml-send-comment.js
0x2b3bfa0: Unavoidable due to monkey-patching of `console.log` as per https://github.com/iterative/cml/pull/581#discussion_r653473819; let's move the discussion there.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: For no specific reason, or so it seems. Unless @DavidGOrtega states otherwise, I would just remove the inner whitespace as you suggest.

bin/cml-send-comment.js
casperdcl: :upside_down_face: lol JS

bin/cml-send-comment.js
casperdcl: :upside_down_face:

src/drivers/github.test.js
0x2b3bfa0: I adjusted them haphazardly to avoid spamming subscribed users with more comments. The values I chose work well enough, but we can run a binary tree search or set in place some profiling to find the optimum values if you want to.

casperdcl: not a massive priority - can revisit if timeouts are actually being hit.

0x2b3bfa0: Agreed! By the way, the Bitbucket timeout is so high because we're running not one but two paginated API queries, each with, literally, a thousand comments from earlier tests. That's the issue with not using ephemeral testing repositories.

src/drivers/bitbucket_cloud.js
DavidGOrtega: As far as I remember markdown needs a space before a new line to actually create a new line? 🤔

casperdcl: did you mean this?

```suggestion
        const longReport = `${commitLink}\n\n${report}`;
```

DavidGOrtega: or this?
```js
const longReport = `${commitLink}\s\n${report}`;
```
I do not remember, I slightly remember that if you remve the space the new line wont happen

0x2b3bfa0: No, he just means that Bitbucket developers... _think differently._

- `hello   \nworld` becomes `<p>hello <br>\nworld</p>`
- `hello  \nworld` becomes `<p>hello<br>\nworld</p>`
- `hello \nworld` becomes `<p>hello \nworld</p>`
- `hello\nworld` becomes `<p>hello\nworld</p>`
- `hello\n\nworld` becomes `<p>hello</p>\n<p>world</p>`
- `hello\n\n\nworld` becomes `<p>hello</p>\n<p>world</p>`

DavidGOrtega: @0x2b3bfa0, Who are you talking to? t @casperdcl? to me?

DavidGOrtega: > Bitbucket developers... think differently.

I can not agree with that more

0x2b3bfa0: Sorry, https://github.com/iterative/cml/pull/581#discussion_r653814301 got stuck and I only saw https://github.com/iterative/cml/pull/581#discussion_r653810810. Now it should be clear.

DavidGOrtega: For the record the Bitbucket markdown does not behave the same and there are some tricks here and there

casperdcl: so... `\n\n` is still the intention, right?

DavidGOrtega: try it and check within Bitbucket @casperdcl

0x2b3bfa0:
Just to spare unnecessary sufferings: that's what I did; check my comment
above.

Using `hello\n\nworld` would produce `<p>hello</p>\n<p>world</p>`

DavidGOrtega (changes requested):
> ⚠️ Bitbucket support won't be fully functional until users (cough) install the Pull Request Commit Links application. In the worst case scenario, commit comments will be created and pull request comments won't.

I wonder if we should unsupport Bitbucket. Sounds very flaky and confusing

src/drivers/bitbucket_cloud.js
0x2b3bfa0: Just to spare unnecessary sufferings: that's what I did; check my comment
above.

Using `hello\n\nworld` would produce `<p>hello</p>\n<p>world</p>`

0x2b3bfa0:
What if we convert the disabled API warning into a fatal error and put it first? This way, users will need to enable the API before using `cml-send-comment` on Bitbucket. Nevertheless, it's as confusing as you think and, even with good documentation, may be challenging for some users.

DavidGOrtega:
> ...it's as confusing as you think and, even with good documentation, may be challenging for some users.

Let's remove it then for Bitbucket.

src/drivers/bitbucket_cloud.js
casperdcl: For avoidance of doubt: I believe

- `\n\n` has and will always, in all flavours of markdown, create a new paragraph
- the intention of the code here was to create a new paragraph

so, er, what's the problem?

casperdcl:
:scream: I wouldn't remove BB functionality just because of a confusing one-click mechanism. Helpful error message (possibly throw/non-zero exit) plz.

src/drivers/bitbucket_cloud.js
0x2b3bfa0: ## `hello␊␊world`

Captura de pantalla 2021-06-18 a las 14 56 54

***

## `hello␠␠␊world`
Captura de pantalla 2021-06-18 a las 14 57 05

0x2b3bfa0: > _so, er, what's the problem?_

None, I think. In fact `␊␊` produces leaves more vertical space than `␠␠␊` and, in my opinion, looks nicer.

0x2b3bfa0: 👍🏼  for replacing it with `\n\n`

DavidGOrtega: I would not say that two lines is nicer

DavidGOrtega:
> 😱 I wouldn't remove BB functionality just because of a confusing one-click mechanism. Helpful error message (possibly throw/non-zero exit) plz.

Ok but also has to be well documented, the last thing I want is we having to answer questions about why is not working in BB

0x2b3bfa0:
@DavidGOrtega & @casperdcl, do you like this fix _à la [#563](https://github.com/iterative/cml/pull/563)ème?_

src/drivers/bitbucket_cloud.js
DavidGOrtega: The intention is one line not  two for the record

casperdcl: Ah so basically we all agree that new paragraph is the intention but:

- I don't care about the amount of vspace, but do prefer universal syntax
- David prefers less vspace by using BB markdown extension syntax
- Helio prefers more vspace by using universal markdown syntax

so in summary we either do

```suggestion
        const longReport = `${commitLink}  \n${report}`;
```

or

```suggestion
        const longReport = `${commitLink}\n\n${report}`;
```

plz pick one. toss a coin if you have to.

DavidGOrtega: It should be only one line, not two

0x2b3bfa0: If Bitcoin block 688142 hash rightmost hex nibble is odd (i.e. 0xf is odd but 0xe is even) I'll agree with @DavidGOrtega. That's how we flip a coin in the 21st century! 🎲 

I would even suggest the solomonic solution of adding the commit hash as a footer instead of a header. See my next comment for a realistic example.

0x2b3bfa0: ## CML Report

| Value   |      Description      |
|----------|:-------------:|
| 42 |  ··· |

***

_Commit comment for 4b799ec1e7673ab716cbd2480308b71942182fdb_

DavidGOrtega: > adding the commit hash as a footer instead of a header. 

We tried that and looked worse and did not have the same feeling as their counterparts. As a header was a better option. Just my two satoshis

0x2b3bfa0: Bitcoin block [688142](https://btc.com/00000000000000000007d08bdea51e4158bc2eab60e1da5413a739220adf05ab) ends in `0xb` (which is odd) so I agree with @DavidGOrtega. 🤷🏼‍♂️  😄

src/drivers/bitbucket_cloud.js
```diff
@@ -22,43 +22,67 @@ class BitBucketCloud {
...
-    const getPrEndpt = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
-    const { values: prs } = await this.request({ endpoint: getPrEndpt });
+    let prs;
+    try {
+      const getPrEndpoint = `/repositories/${projectPath}/commit/${commitSha}/pullrequests`;
+      prs = await this.paginatedRequest({ endpoint: getPrEndpoint });
+    } catch (err) {
+      if (err.message === 'Not Found Resource not found')
+        err.message =
+          "Click 'Go to pull request' on any commit details page to enable this API";
```
casperdcl, line 35: maybe also add a link to the relevant BB docs which state this (for inquisitive user reference)?

0x2b3bfa0: @DavidGOrtega seems to dislike errors with links, but I actually find them an awesome resource on modern terminal emulators.

Even if it's out of scope for this pull request, I would love to see links like `https://err[or].cml.dev/YpQ` on those errors that are especially tricky to explain inline or require additional context. If I recall correctly, DVC does something similar.

casperdcl: error message with link prevents a user opening an issue "I got this error saying I haz to click somewhere can u plz automate it"

0x2b3bfa0: Moved to https://github.com/iterative/cml/issues/606

casperdcl (approved):

0x2b3bfa0:
I miss #538 with granular tests and ephemeral repositories. 🐲  Having to manually delete comments before running automated tests is... far from ideal.

```console
FAIL src/drivers/bitbucket_cloud.test.js
[Forbidden] By default, you can't create more than 1000 comments per pull request. Contact support to update this limit.
```

0x2b3bfa0:
@DavidGOrtega, may I have your explicit approval so the pull request gets merged with everything green?

DavidGOrtega:
> I miss #538 with granular tests and ephemeral repositories. 🐲 Having to manually delete comments before running automated tests is... far from ideal.

FAIL src/drivers/bitbucket_cloud.test.js
[Forbidden] By default, you can't create more than 1000 comments per pull request. Contact support to update this

This is weird, we fixed this already. As far as I remmber

PR #584, opened 2021-06-10
ssh_private is valid in TF

DavidGOrtega:
We are masking a non correct member. Json is generated by Terraform and the value is ssh_private


0x2b3bfa0 (approved):
🙈  🙏🏼  💯

casperdcl (approved):
groan

PR #585, opened 2021-06-10
Runners not unregistering due to camelCase

DavidGOrtega:
GH api uses snake_case 🐍  due to this GH are not unregistering successfully anymore

Fixed in #583 however it might take longer and this is important.

### DIFF
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -144,18 +144,18 @@ class Github {
     const { name } = opts;
     const { owner, repo } = ownerRepo({ uri: this.repo });
     const { actions } = octokit(this.token, this.repo);
-    const { id: runner_id } = await this.runnerByName({ name });
+    const { id } = await this.runnerByName({ name });
 
     if (typeof repo !== 'undefined') {
       await actions.deleteSelfHostedRunnerFromRepo({
         owner,
         repo,
-        runner_id
+        runner_id: id
       });
     } else {
       await actions.deleteSelfHostedRunnerFromOrg({
         org: owner,
-        runner_id
+        runner_id: id
       });
     }
   }
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -144,18 +144,18 @@ class Github {
     const { name } = opts;
     const { owner, repo } = ownerRepo({ uri: this.repo });
     const { actions } = octokit(this.token, this.repo);
-    const { id: runnerId } = await this.runnerByName({ name });
+    const { id: runner_id } = await this.runnerByName({ name });
 
     if (typeof repo !== 'undefined') {
       await actions.deleteSelfHostedRunnerFromRepo({
         owner,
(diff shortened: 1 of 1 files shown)


src/drivers/github.js
```diff
@@ -144,18 +144,18 @@ class Github {
...
         owner,
         repo,
-        runnerId
+        runner_id
       });
     } else {
       await actions.deleteSelfHostedRunnerFromOrg({
         org: owner,
-        runnerId
+        runner_id
```
0x2b3bfa0, line 158:
```suggestion
        runner_id: runnerId
```

src/drivers/github.js
```diff
@@ -144,18 +144,18 @@ class Github {
...
     const { actions } = octokit(this.token, this.repo);
-    const { id: runnerId } = await this.runnerByName({ name });
+    const { id: runner_id } = await this.runnerByName({ name });
 
     if (typeof repo !== 'undefined') {
       await actions.deleteSelfHostedRunnerFromRepo({
         owner,
         repo,
-        runnerId
+        runner_id
```
0x2b3bfa0, line 153:
```suggestion
        runner_id: runnerId
```

src/drivers/github.js
```diff
@@ -144,18 +144,18 @@ class Github {
     const { name } = opts;
     const { owner, repo } = ownerRepo({ uri: this.repo });
     const { actions } = octokit(this.token, this.repo);
-    const { id: runnerId } = await this.runnerByName({ name });
+    const { id: runner_id } = await this.runnerByName({ name });
```
0x2b3bfa0, line 147:
```suggestion
    const { id: runnerId } = await this.runnerByName({ name });
```

0x2b3bfa0 (approved):
What if we delay the conversion to snake_case until the invocation of the GitHub API? Anyhow, it might be a bit more verbose, so I don't have a strong preference over any of the methods.

0x2b3bfa0 (approved):
Looks good to me!

DavidGOrtega:
> What if we delay the conversion to snake_case until the invocation of the GitHub API? Anyhow, it might be a bit more verbose, so I don't have a strong preference over any of the methods.

Its ugly, normally in modern js you destructure looking to match the needed member later on. In this case Im going to reuse id straight to avoid lint but in the future we should destructure snake_case and add a lint ignore rule

casperdcl (commented):
I also prefer the "ugly but correct/consistent" way suggested by @0x2b3bfa0 and used in the rest of the codebase (and my Python bias avoids variables called `id`) but whatever

DavidGOrtega:
In this case was done just because is easy to read. If it were 20 lines I would not have done it either. But here is totally fine because you use it afterwards assignment

DavidGOrtega:
unless your memory is worse than mine 👴

PR #540, opened 2021-05-19
Rename --cloud-gpu tesla to v100

0x2b3bfa0:
Fixes iterative/cml#509 in combination with iterative/terraform-provider-iterative#112

⚠️ **Important note:** to be released after iterative/terraform-provider-iterative#112

PR #602, opened 2021-06-18
Allow newlines in SSH private keys

0x2b3bfa0:
This pull request simplifies the SSH key format requirements so users can provide keys directly without having to escape newlines on their own.

#### Before
```console
$ cml-runner --cloud-ssh-private="$(perl -pe 's/\n/\\n/g' key.pem)"
```

#### After
```console
$ cml-runner --cloud-ssh-private="$(cat key.pem)"
```

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -379,7 +379,7 @@ const opts = yargs
     'cloud-ssh-private',
     'Custom private RSA SSH key. If not provided an automatically generated throwaway key will be used'
   )
-  .coerce('cloud-ssh-private', (val) => val.replace(/\n/g, '\\n')
+  .coerce('cloud-ssh-private', (val) => val.replace(/\n/g, '\\n'))
   .boolean('cloud-ssh-private-visible')
   .describe(
     'cloud-ssh-private-visible',
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -163,7 +163,7 @@ const runCloud = async (opts) => {
         type,
         gpu: gpu === 'tesla' ? 'v100' : gpu,
         hddSize,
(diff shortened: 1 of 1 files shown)


bin/cml-runner.js
```diff
@@ -163,7 +163,7 @@ const runCloud = async (opts) => {
         type,
         gpu: gpu === 'tesla' ? 'v100' : gpu,
         hddSize,
-        sshPrivate,
+        sshPrivate.replace(/\n/g, '\\n'),
```
DavidGOrtega, line 166: Better if we make it in a coerce at yargs? We are now using it and Im also adding it in #583 

```js
.default('cloud-ssh-private', '')
  .describe(
    'cloud-ssh-private',
    'Custom private RSA SSH key. If not provided an automatically generated throwaway key will be used'
  )
.coerce('cloud-ssh-private', val => val.replace(/\n/g, '\\n')
```

DavidGOrtega (changes requested):

bin/cml-runner.js
0x2b3bfa0: Yes, far better. In fact, we should probably do the same with the `gpu` aliases.

DavidGOrtega:
```js
.coerce('cloud-gpu-type', val => val === 'nogpu' ? null : val)
```

We do it with cloud-gpu-type but Im not sure what is gpu aliases

0x2b3bfa0: I meant the change we introduced with https://github.com/iterative/cml/pull/540; using `coerce` could also be better.

DavidGOrtega (approved):
👍 🚀

casperdcl (approved):

Issue #597, opened 2021-06-16
Drop support for Node.js 12 or use a transpiler


0x2b3bfa0:
Node.js 12 is [in maintenance mode](https://nodejs.org/en/about/releases/) and will stop being supported on 2022-04-30. It would be nice to start using, Node.js 14 or even 16 to take advantage of features like [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) as per https://github.com/iterative/cml/pull/581#discussion_r652756206.

DavidGOrtega:
I don't have to wait for a specific node version we can always use a transpiler.

0x2b3bfa0:
If we choose that path, it would be nice to consider TypeScript. Anyhow, we
can remain in the pure JavaScript realms with Babel.

0x2b3bfa0:
See also https://github.com/iterative/setup-cml/issues/40

0x2b3bfa0:
Already using optional chaining:
* https://github.com/iterative/cml/pull/943#pullrequestreview-937334724

Issue #596, opened 2021-06-16
Remove Restyle in CML or adjust it to follow airbnb style


DavidGOrtega:
Having some clear difficulties in having consistency in style, its better to just let the linter speak and refuse the PR if the style does not match the chosen style.

We have to:
 - Remove Restyle or make it use our eslint config (I remember that to not be possible)

This will make contributions harder but at least the base source code will follow the same convention.

0x2b3bfa0:
Existential doubt from https://github.com/iterative/cml/pull/581#discussion_r652600023:

```javascript
(item) => {
  const { property: { data = '' } = {} } = item;
  return data.endsWith('text');
}
```

_versus_

```javascript
(item) => item && item.property && item.property.data && item.property.data.endsWith('text')
```

_versus_

```javascript
(item) => item?.property?.data?.endsWith('text')
```

The last one is only valid in Node.js 14+ as stated on #597

0x2b3bfa0:
Note: as per https://github.com/iterative/cml/pull/581#discussion_r652781670, we probably should broaden a bit the scope of the issue title.

0x2b3bfa0:
Another style topic: https://github.com/iterative/cml/pull/583#discussion_r657433898

0x2b3bfa0:
Speaking about what linters can't detect: how deffensive should we be when writing code that interacts with complex systems out of our control?

- https://github.com/iterative/cml/pull/468#discussion_r610534623
- https://github.com/iterative/cml/pull/563#issue-654270569
- https://github.com/iterative/cml/pull/583#discussion_r657449080
- https://github.com/iterative/cml/pull/608#issuecomment-868690031

0x2b3bfa0:
Already using optional chaining 🎉 
* https://github.com/iterative/cml/pull/943#pullrequestreview-937334724

0x2b3bfa0:
Tracking https://github.com/iterative/cml/pull/1026#discussion_r894820360

PR #608, opened 2021-06-25
Fix endpoint detection logic, broken on FetchError

0x2b3bfa0:
Detecting errors other than the vanilla `Error` isn't especially easy nor reliable, or so it seems. Given that we only return errors and strings, it's easier to consider as an error everything that isn't a string.

### DIFF
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -60,7 +60,7 @@ class Gitlab {
     );
 
     this.detectedBase = possibleBases.find(
-      (base) => base.constructor !== Error
+      (base) => typeof base === 'string'
     );
     if (!this.detectedBase) {
       if (possibleBases.length) throw possibleBases[0];
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -59,9 +59,7 @@ class Gitlab {
         })
     );
 
-    this.detectedBase = possibleBases.find(
-      (base) => typeof base === 'string'
-    );
+    this.detectedBase = possibleBases.find((base) => typeof base === 'string');
     if (!this.detectedBase) {
       if (possibleBases.length) throw possibleBases[0];
       throw new Error('Invalid repository address');


0x2b3bfa0:
Spent 30+ minutes figuring out why [this user](https://discord.com/channels/485586884165107732/728693131557732403/858002229348663306) was querying a project with a broken full address beginning with `ttps` instead of the usual `user/project` format.

It turns out that the [`projectPath`](https://github.com/iterative/cml/blob/cf0ba10e4383315b0cb406f74f528a3c6ebec896/src/drivers/gitlab.js#L30-L37) function wasn't checking if `repo.startsWith(repoBase)` before running `repo.replace(repoBase, '').substr(1)` and it just worsened the issue by removing the first character without removing the expected prefix.

Are my deffensive programming instincts worth the time? Not sure, but... https://github.com/iterative/cml/pull/468#discussion_r610540799

DavidGOrtega:
Not sure if this is solving the real issue. And of course nothing to be with the original case. The problem is that in CML.js we assume always ```https```.

0x2b3bfa0:
> _Not sure if this is solving the real issue._

This pull request just fixes a bug where every connection error gets masked under a barely meaningful `TypeError: Only absolute URLs are supported` error. Not sure if that's the _real issue_ you mention, but at least it helps to give unerring replies to users with _real issues_ reporting the mentioned unhelpful error.

> _And of course nothing to be with the original case._

If you mean [this Discord conversation](https://discord.com/channels/485586884165107732/728693131557732403/858002229348663306), it would have been impossible to provide an accurate solution without the code from this pull request. 🤔 

> _The problem is that in CML.js we assume always https._

That's a good default, but some self-hosted users might still be relying on bare HTTP. It looks like you've identified another potential issue.

DavidGOrtega:
> That's a good default, but some self-hosted users might still be relying on bare HTTP. It looks like you've identified another potential issue.

I think thats the real issue of your user.

0x2b3bfa0:
> _I think thats the real issue of your user._

It looks like "my" user had [a slightly more creative issue](https://discord.com/channels/485586884165107732/728693131557732403/858027874966437938), where a plain HTTP server was listening on port 443 instead of 80.

> _All this pain could be solved checking http vs https since http wont even work either in any circumstance_

How come?

src/drivers/gitlab.js
```diff
@@ -59,9 +59,7 @@ class Gitlab {
         })
     );
 
-    this.detectedBase = possibleBases.find(
-      (base) => base.constructor !== Error
-    );
+    this.detectedBase = possibleBases.find((base) => typeof base === 'string');
```
DavidGOrtega, line 62: Could we use?

```js
filter(base => base.startsWith('http'))
```

0x2b3bfa0: Abracadabra? https://github.com/iterative/cml/issues/596#issuecomment-862491282 🧙🏼‍♂️ 

```js
filter(base => base && base.startsWith && base.startsWith('http'))
```

After #597 it could even be written like this:

```js
filter(base => base?.startsWith?.('http'))
```

DavidGOrtega: even better

```js
filter(base => typeof base === 'string'  && base.startsWith('http'))
```

0x2b3bfa0: > _Could we use?_
> 
> ```js
> filter(base => base && base.startsWith && base.startsWith('http'))
> ```

Assuming that you mean `find` and not `filter` as the used method, yes. Your suggestion seems to cover every case I can think of, even if the intent is slightly different.

DavidGOrtega:
```js
find(base => typeof base === 'string'  && base.startsWith('http'))
```

0x2b3bfa0: If we check **begins with `http`** instead of **is not an error**, we would be introducing an additional constraint. Even if it seems apparently harmless and conveys a more explicit intention, I don't know if we might be hitting edge cases where it would fail, like `Https` (note the uppercase `H` character).

DavidGOrtega:
```js
find(base => typeof base === 'string'  && base.toLowerCase().startsWith('http'))
```

0x2b3bfa0: We only return a string after receiving a well-formed API response, and we shouldn't need to check again on our own whether the protocol is HTTP or not.

https://github.com/iterative/cml/blob/c9b006038c34023eef837c4d995223c0c22a6dc8/src/drivers/gitlab.js#L50-L58

0x2b3bfa0: Anyhow, if you consider that there is a valid reason to add this additional check, I'm open to it.

DavidGOrtega: nope, its totally wrong. they are pathnames, they dont even start with any protocol.

DavidGOrtega: I do not fully understand how you got ```ttp``` there

DavidGOrtega: I still think that the error is something related to http vs https maybe?

if they give you a repo like 'http://something' we are going to hit that 'ttp' because of the assumed ```https``` in CML.js and the ```replace().substr(1)```

0x2b3bfa0: > _nope, it's totally wrong. they are pathnames, they don't even start with any protocol._

Each string on the `possibleBases` array is a full URL that's proven to be a GitLab API base, and all of them do start with a protocol like `http` or `https` as you can see by debugging the code.

https://github.com/iterative/cml/blob/c9b006038c34023eef837c4d995223c0c22a6dc8/src/drivers/gitlab.js#L43

> _I do not fully understand how you got `ttp` there_
> _if they give you a repo like 'http://something' we are going to hit that 'ttp' because of the assumed `https` in CML.js and the `replace().substr(1)` ?_

Before this pull request, the method to get the first valid API base address from `possibleBases` was the following:

https://github.com/iterative/cml/blob/6c451d965aaea2ad2b1812dd2a223def9ce87739/src/drivers/gitlab.js#L62-L64

Unfortunately, when `base` is a `FetchError` object, the `base.constructor !== Error` check doesn't have the desired effect and `this.detectedBase` will get assigned to the first `FetchError` on the `possibleBases` array.

https://github.com/iterative/cml/blob/6c451d965aaea2ad2b1812dd2a223def9ce87739/src/drivers/gitlab.js#L30-L37

Trying to replace `FetchError: a long description of the error` by an empty string on the code above would leave the original `this.repo` string intact. Then, you remove the first character of the result with `.substr(1)` so `this.repo` loses its first character.

0x2b3bfa0: > _I still think that the error is something related to http vs https maybe?_

Users may have any connection error, and all of them will be masked under an unhelpful `TypeError: Only absolute URLs are supported` until we merge this pull request. HTTP versus HTTPS errors fall in this category.

DavidGOrtega: > TypeError: Only absolute URLs are supported

Is this PR improving a better message 😃

0x2b3bfa0: > _Is this PR improving a better message_

It doesn't improve any message _per se,_ it just avoids masking a helpful error behind an unhelpful one.

PR #613, opened 2021-06-28
Skip Terraform version checks when not needed

### LINKED ISSUE #612: Terraform dependency should be only a must if the resource is set
```sh
cml-runner --repo YYY --token XXX--labels gpu --idle-timeout 0 --retry
{"level":"error","status":"terminated"}
Error: terraform version

        /bin/sh: 1: terraform: not found
```

0x2b3bfa0:
Terraform should not be a dependency for invocations that don't include `--cloud` or `--tf-resource` options. Should fix #612.

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -39,7 +39,7 @@ const shutdown = async (opts) => {
   RUNNER_SHUTTING_DOWN = true;
 
   let { error, cloud } = opts;
-  const { name, workdir = '' } = opts;
+  const { name, tfResource, workdir = '' } = opts;
   const tfPath = workdir;
 
   console.log(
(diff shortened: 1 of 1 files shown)


DavidGOrtega (changes requested):
It's not enough. It should be check also in the termination

0x2b3bfa0:
@DavidGOrtega, is there any scenario where the termination code is executed without reaching first the startup code?

DavidGOrtega:
Non cloud runner do not have tf-resource however when terminating is going to try to destroy

0x2b3bfa0:
🤦🏼  So true!

0x2b3bfa0:
@DavidGOrtega, please take a look at the latest changes. I haven't tested them yet, but it looks like that should do the trick. 🙊

DavidGOrtega (approved):
👍 lgtm however I will try it first. You dont have to!

PR #626, opened 2021-06-30
Require runners to be online when honoring --reuse

0x2b3bfa0:
The `--reuse` option should not consider offline runners as reusable.

### DIFF
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -225,19 +225,19 @@ class Github {
     const { actions } = octokit(this.token, this.repo);
     let runners = [];
 
-    if (typeof repo !== 'undefined') {
+    if (typeof repo === 'undefined') {
       ({
         data: { runners }
-      } = await actions.listSelfHostedRunnersForRepo({
-        owner,
-        repo,
+      } = await actions.listSelfHostedRunnersForOrg({
+        org: owner,
         per_page: 100
       }));
     } else {
       ({
         data: { runners }
-      } = await actions.listSelfHostedRunnersForOrg({
-        org: owner,
+      } = await actions.listSelfHostedRunnersForRepo({
+        owner,
+        repo,
         per_page: 100
       }));
     }
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -222,59 +222,27 @@ class Github {
 
   async getRunners(opts = {}) {
     const { owner, repo } = ownerRepo({ uri: this.repo });
-    const { actions } = octokit(this.token, this.repo);
-    let runners = [];
+    const { paginate, actions } = octokit(this.token, this.repo);
 
+    let runners;
     if (typeof repo === 'undefined') {
-      ({
-        data: { runners }
-      } = await actions.listSelfHostedRunnersForOrg({
-        org: owner,
-        per_page: 100
-      }));
+      runners = await paginate(actions.listSelfHostedRunnersForOrg, {
+        org: owner
+      });
     } else {
-      ({
-        data: { runners }
-      } = await actions.listSelfHostedRunnersForRepo({
+      runners = await paginate(actions.listSelfHostedRunnersForRepo, {
         owner,
-        repo,
-        per_page: 100
-      }));
+        repo
+      });
     }
 
-    return runners;
-  }
-
-  async runnerByName(opts = {}) {
-    const { name } = opts;
-    const runners = await this.getRunners(opts);
-    const runner = runners.find((runner) => runner.name === name);
-    if (runner)
-      return {
-        id: runner.id,
-        name: runner.name,
-        busy: runner.busy,
-        online: runner.status === 'online'
-      };
-  }
-
-  async runnersByLabels(opts = {}) {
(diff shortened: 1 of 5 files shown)


src/drivers/github.js
```diff
@@ -225,19 +225,19 @@ class Github {
     const { actions } = octokit(this.token, this.repo);
     let runners = [];
 
-    if (typeof repo !== 'undefined') {
+    if (typeof repo === 'undefined') {
```
DavidGOrtega, line 228: Why this change? something related to your non OCD? 😄

src/drivers/gitlab.js
```diff
@@ -181,18 +181,29 @@ class Gitlab {
...
     const endpoint = `/runners?per_page=100`;
     const runners = await this.request({ endpoint, method: 'GET' });
-    const runner = runners.filter(
+    const runner = runners.find(
       (runner) => runner.name === name || runner.description === name
-    )[0];
+    );
 
-    if (runner) return { id: runner.id, name: runner.name };
+    if (runner)
```
DavidGOrtega, line 188: This could be dryer.

I would do:

```js
async runnersByLabels(opts = {}) {
    const { labels } = opts;
    const endpoint = `/runners?per_page=100&${labels ? `tag_list=${labels}`: ''}`;
    const runners = await this.request({ endpoint, method: 'GET' });
    return runners.map((runner) => ({
      id: runner.id,
      name: runner.name,
      busy: runner.active,
      online: runner.status === 'online'
    }));
  }

async runnerByName(opts = {}) {
    const { name } = opts;
    const runners = await this.runnersByLabels();
    return runners.find(
      (runner) => runner.name === name || runner.description === name
    );
  }
```

RunnersbyName might be something very specific and a bit misleading. Feel free to do a runners method or as I did (to not break sync between drivers) we can assume that if you do not specify labels is returning all of them

src/drivers/github.js
```diff
@@ -248,8 +248,14 @@ class Github {
   async runnerByName(opts = {}) {
```
DavidGOrtega, line 248: Dryer like below.

src/drivers/github.js
0x2b3bfa0: Yes, actually a Pythonic "bad habit" of using posssitive conditions whenever possible. You can count it as part of the OCD I «don't not have» (note the double negation). 😄  Feel free to revert if you don't think it matters.

DavidGOrtega: Its fine I support every OCD habit since I have many

src/drivers/gitlab.js
DavidGOrtega:
```
(runner) => runner.name === name || runner.description === name
```

should be

```
(runner) => runner.name === name 
```

Since description is not in the data model anymore

src/drivers/github.js
0x2b3bfa0: What is _below?_ 👀

DavidGOrtega: above!! For some reason I thought that I was commenting the chunk on top

src/drivers/gitlab.js
0x2b3bfa0: > _RunnersbyName might be something very specific and a bit misleading_

Actually it's `runnerByName` without the plural, that's why I kept it: `runnersByLabels` returns runner**s** that contain all the specified labels, and `runnerByName` returns the **only** runner that has the specified unique name.

0x2b3bfa0: You're going to laugh with this: actually, the `description` field is the only holding the name, and the `name` field is `null` in all the cases I could test.

0x2b3bfa0: I've chosen to make `getRunners` universal on the driver and extract both `runnerByName` and `runnersByLabels` to `cml.js`

What do you think?

DavidGOrtega: Gitlab used description as far as I remember

DavidGOrtega: we treated description AS name

src/drivers/gitlab.js
```diff
@@ -176,23 +176,20 @@ class Gitlab {
...
-
-  async runnersByLabels(opts = {}) {
-    const { labels } = opts;
-    const endpoint = `/runners?per_page=100?tag_list=${labels}`;
-    const runners = await this.request({ endpoint, method: 'GET' });
-    return runners.map((runner) => ({ id: runner.id, name: runner.name }));
+    return await Promise.all(
+      runners.map(async ({ id, name, description, active, online }) => ({
+        id,
+        name: name || description,
```
0x2b3bfa0, line 185:
```suggestion
        name: description,
```

The `name` property is always `null` because [`POST /runners`](https://docs.gitlab.com/ee/api/runners.html#register-a-new-runner) doesn't accept it. We're using `description` instead of name to store a custom string.

DavidGOrtega: I would say this is true in the meantime GL fixes this bug on their side. The name should be changed and not the description so I would say is ok for now

PR #648, opened 2021-07-07
Fix unhandled promise rejection on tensorboard-dev.js

### LINKED ISSUE #479: cml-tensorboard-dev bad credentials give a non descriptive error
It returns an unhandled promise

### LINKED ISSUE #647: Infinite TB loop
Somehow TB can loop forever instead of just failing. This is terrible for cloud instances.

```sh
(node:1809) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of null
    at Timeout._onTimeout (/usr/lib/node_modules/@dvcorg/cml/bin/cml-tensorboard-dev.js:71:17)
(node:1809) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:1809) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:1809) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of null
    at Timeout._onTimeout (/usr/lib/node_modules/@dvcorg/cml/bin/cml-tensorboard-dev.js:71:17)
(node:1809) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:1809) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of null
    at Timeout._onTimeout (/usr/lib/node_modules/@dvcorg/cml/bin/cml-tensorboard-dev.js:71:17)
(node:1809) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 3)
```

0x2b3bfa0:
@0x2b3bfa0  I have taken this PR

 - closes #647 
 - closes #479

### DIFF
--- a/bin/cml-tensorboard-dev.js
+++ b/bin/cml-tensorboard-dev.js
@@ -68,7 +68,7 @@ const run = async (opts) => {
     const regex = /(https?:\/\/[^\s]+)/;
     const matches = stdoutData.match(regex);
 
-    if (matches.length) {
+    if (matches && matches.length) {
       let output = matches[0];
 
       if (!rmWatermark) output = watermarkUri({ uri: output, type: 'tb' });
--- a/bin/cml-tensorboard-dev.js
+++ b/bin/cml-tensorboard-dev.js
@@ -12,6 +12,7 @@ const tempy = require('tempy');
 const { exec, watermarkUri } = require('../src/utils');
 
 const { TB_CREDENTIALS } = process.env;
+const isCLI = require.main === module;
 
 const closeFd = (fd) => {
   try {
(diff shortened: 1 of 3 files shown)


bin/cml-tensorboard-dev.js
```diff
@@ -68,7 +68,7 @@ const run = async (opts) => {
     const regex = /(https?:\/\/[^\s]+)/;
     const matches = stdoutData.match(regex);
 
-    if (matches.length) {
+    if (matches && matches.length) {
```
DavidGOrtega, line 71: This could definitely solve the error... but the loop? I think that there is a retry in a loop to endup with that result

DavidGOrtega: Indeed the error has to be that the output is not what is expected

0x2b3bfa0: The loop will break after 60 seconds of invalid results. I think that it's the expected behavior?

0x2b3bfa0: This was broken because `"hello".match(/(world)/) === null` and it errored each time it didn't find a valid link.

DavidGOrtega: The question is... why it should not receive the link?

0x2b3bfa0: Because of some error we're missing due to the lack of debug output? If you're able to reproduce this issue consistently, it would be nice to take a look at the full standard output/error.

DavidGOrtega:
@0x2b3bfa0 I have taken this PR but you can not review it now

PR #651, opened 2021-07-09
GL restarts the whole pipeline

DavidGOrtega:
- GL restart whole pipeline. GL pipeline retry only retries failed jobs within a pipeline and not the whole pipeline 🤦


0x2b3bfa0 (approved):
Looks good to me (static)

PR #653, opened 2021-07-11
Cloud runner terminate improvements

### LINKED ISSUE #407: cml-runner - Better idle mechanism in GH
GH runners sometimes hangs when you cancel the job staying idle and not destroying themselves.
Probably for not finishing the current job not allowing idle to come back.

### LINKED ISSUE #649: Instances intermittently fail to terminate
I've had a couple of instances recently that have failed to terminate. In the most recent case this was with the `--reuse` flag set, having run a series of 8 queued jobs.

The instance is sitting idle, with a timeout of `60s` having passed ten minutes ago. I'll need to terminate the instance manually from the command line.

In the most serious case, I had an instance run for two weeks without terminating. It took so long for us to notice because the instance name did not get set to `cml-*` as usual.

Here's the yml we are using:

```
name: train and evaluate rasa model

on:
  pull_request:
    types: [opened, synchronize]
  workflow_dispatch:

jobs:
  deploy-runner:
    runs-on: [ubuntu-latest]
    steps:
      - uses: actions/checkout@v2
      - uses: iterative/setup-cml@v1

      - name: deploy
        shell: bash
        env:
          REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
        run: |
          cml-runner \
          --cloud aws \
          --cloud-region eu-west \
          --cloud-type=c5a.4xlarge \
          --cloud-spot true \
          --labels=cml-runner,voice-control,oms-rasa-2 \
          --idle-timeout 60 \
          --reuse
  model-training:
    needs: deploy-runner
    runs-on: [self-hosted,cml-runner]
    container: docker://dvcorg/cml:0-dvc2-base1

    steps:
    - uses: actions/checkout@v2
      with: 
        ref: ${{ github.event.pull_request.head.sha }}

      with:
        python-version: '3.8.5'
    - name: Install dependencies
      run: |
        apt-get update -y
        apt-get install make python3-pip virtualenv curl
    - name: cml
      env:
        REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
        AWS_REGION: eu-west-1
      run: |
        python --version
        make virtualenv
        dvc repro
        echo "## Metrics" > report.md
        git fetch --prune
        dvc metrics diff main --show-md | grep "Change\|\-\-\-" >> report.md
        dvc metrics diff main --show-md | grep -E "(intent|entity|action).*weighted" | sort >> report.md
        sed "s/results\///g" -i report.md
        cml-send-comment report.md
        dvc push

    - uses: actions/upload-artifact@v2
      with:
        name: gh-artifact-${{ github.event.pull_request.head.sha }}
        path: |
          report.md
          results
        retention-days: 30
        
    - uses: EndBug/add-and-commit@v7
      if: ${{ github.ref != 'refs/heads/main' }} && ${{ github.ref != 'refs/heads/rasax/prod' }}
      with:
         add: 'dvc.lock --force'
         pull_strategy: 'NO-PULL'
         message: 'chg: dvc repro'

### LINKED ISSUE #650: CML seemingly fails to restart job after AWS Spot instances have been shut down
Hey everyone, 
So I noticed a couple of days ago that CML now has new functionality that allows it to restart workflows if one or more AWS spot runners have been told to shut down. However this doesn't seem to be happening for me. 

A couple of details about our case: 
* Our cloud is AWS 
* We're (as far as i can tell) using the `latest` version of CML to deploy a bunch of runners as shown below. 
```
  deploy_runners:
    name: Deploy Cloud Instances
    needs: [setup_config]

    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2
      - uses: iterative/setup-cml@v1
        with:
          version:  latest

      - name: "Deploy runner on EC2"
        shell: bash
        env:
          repo_token: ${{ secrets.ACCESS_TOKEN_CML }}
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_TESTING }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_TESTING }}
          CASE_NAME: ${{ matrix.case_name }}
          N_RUNNERS: ${{ fromJson(needs.setup_config.outputs.json_string).n_runners }}

        run: |
          for (( i=1; i<=N_RUNNERS; i++ ))
          do
            echo "Deploying runner ${i}"
            cml-runner \
            --cloud aws \
            --cloud-region eu-west-2 \
            --cloud-type=m \
            --cloud-hdd-size 100 \
            --cloud-spot \
            --labels=cml-runner &
          done
          wait
          echo "Deployed ${N_RUNNERS} runners."
```

* The job each runner runs does not use the CML images provided by iterative 
* The job that each runner runs has `continue-on-error` set to `False` (wondering whether that is interfering with cml?) 
```
  run_optimisation:
    continue-on-error: false
    strategy:
      matrix: ${{fromJson(needs.setup_config.outputs.json_string).matrix}}
      fail-fast: true

    runs-on: [self-hosted, "cml-runner"]
    container:
      image: python:3.8.10-slim
      volumes:
          - /dev/shm:/dev/shm
```

### LINKED ISSUE #661: GCP compute instances are not shutdown after idle timeout
First off, thanks for adding GCP support to cml-runner.

While playing around with it, I noticed that my compute engine instances weren't being shutdown/terminated (i.e., VM powered off) or deleted. I experimented with `--idle-timeout` and `--single`, yet neither made a difference. The instance stays alive. This is unexpected given [the following sentence from the docs](https://cml.dev/doc/self-hosted-runners#:~:text=After%20the%20job%20runs%2C%20the%20instance%20automatically%20shuts%20down.) on self-hosted runners:

> After the job runs, the instance automatically shuts down.

And indeed, that's what I've observed with AWS instances. Those seem to terminate correctly.

Here's my Github workflow for extra context:

```
name: 'Train-in-the-cloud-GCP'
on: 
  workflow_dispatch:

jobs:
  deploy-runner:
    runs-on: [ubuntu-latest]
    steps:
      - uses: iterative/setup-cml@v1
      - uses: actions/checkout@v2
      - name: 'Deploy runner on GCP'
        shell: bash
        env:
          REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
          # Notice use of `GOOGLE_APPLICATION_CREDENTIALS_DATA` instead of
          # `GOOGLE_APPLICATION_CREDENTIALS`. Contrary to what docs suggest, the
          # latter causes problems for terraform.
          GOOGLE_APPLICATION_CREDENTIALS_DATA: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
        run: |
          cml-runner \
          --cloud gcp \
          --cloud-region europe-west1-b	 \
          --cloud-type=n1-standard-1 \
          --labels=cml-runner
          
  model-training:
    needs: deploy-runner
    runs-on: [self-hosted, cml-runner]
    container: docker://dvcorg/cml-py3:latest
    steps:
      - uses: actions/checkout@v2
      - name: 'Train my dummy model'
        env:
          REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
        run: |
          echo "Training a super awesome model"
          sleep 5
          echo "Training complete"
```

Hope there's a way to ensure the same auto-shutdown behavior on GCP. As it is, the risk of getting smacked with an expensive bill for an idle GPU is just too real =p

### LINKED ISSUE #662: Single accepts only one job but does not cleanup
And strangely appears as offline.

image

### PR DESCRIPTION
**Pull request by DavidGOrtega**
- Introduces waitCompletedPipelineJobs a function that waits for expected completed jobs. fixes #649 
  - Adds proc on exit event. fixes #662. closes #407 
  - After testing. closes #661 
  - SpotNotifier. Closes #650

### DIFF
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -251,13 +251,11 @@ const runLocal = async (opts) => {
   proc.stderr.on('data', dataHandler);
   proc.stdout.on('data', dataHandler);
   proc.on('uncaughtException', () => shutdown(opts));
-  proc.on('SIGINT', () => shutdown(opts));
-  proc.on('SIGTERM', () => shutdown(opts));
-  proc.on('SIGQUIT', () => shutdown(opts));
+  proc.on('disconnect', () => shutdown(opts));
 
-  if (parseInt(idleTimeout) !== 0) {
+  if (cml.driver === 'github' && parseInt(idleTimeout) !== 0) {
     const watcher = setInterval(() => {
-      RUNNER_TIMEOUT_TIMER >= idleTimeout &&
+      RUNNER_TIMEOUT_TIMER > idleTimeout &&
         shutdown(opts) &&
         clearInterval(watcher);
 
@@ -265,8 +263,8 @@ const runLocal = async (opts) => {
     }, 1000);
   }
 
-  if (!noRetry && cml.driver === 'github') {
-    const watcher = setInterval(() => {
+  if (cml.driver === 'github' && !noRetry) {
+    const watcher = setInterval(async () => {
       RUNNER_JOBS_RUNNING.forEach((job) => {
         if (
           new Date().getTime() - new Date(job.date).getTime() >
@@ -274,6 +272,11 @@ const runLocal = async (opts) => {
         )
           shutdown(opts) && clearInterval(watcher);
       });
+
+      const runner = await cml.runnerByName({ name: name });
+      if (runner && runner.status && runner.status.toLowerCase() === 'idle') {
+        RUNNER_JOBS_RUNNING = [];
+      }
     }, 60 * 1000);
   }
 
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -215,7 +215,7 @@ const runCloud = async (opts) => {
 
 const runLocal = async (opts) => {
   console.log(`Launching ${cml.driver} runner`);
-  const { workdir, name, labels, single, idleTimeout, noRetry, cloud } = opts;
+  const { workdir, name, labels, single, idleTimeout, noRetry } = opts;
 
   const proc = await cml.startRunner({
     workdir,
@@ -277,10 +277,12 @@ const runLocal = async (opts) => {
   proc.on('disconnect', () => shutdown(opts));
   proc.on('exit', () => shutdown(opts));
 
-  if (cloud === 'aws') {
-    await SpotNotifier.instanceId();
+  try {
+    console.log(`EC2 id ${await SpotNotifier.instanceId()}`);
     SpotNotifier.on('termination', () => shutdown(opts));
     SpotNotifier.start();
+  } catch (err) {
+    console.log('SpotNotifier can not be started.');
   }
 
   if (parseInt(idleTimeout) !== 0) {
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -277,12 +277,14 @@ const runLocal = async (opts) => {
   proc.on('disconnect', () => shutdown(opts));
   proc.on('exit', () => shutdown(opts));
 
-  try {
-    console.log(`EC2 id ${await SpotNotifier.instanceId()}`);
-    SpotNotifier.on('termination', () => shutdown(opts));
-    SpotNotifier.start();
-  } catch (err) {
-    console.log('SpotNotifier can not be started.');
+  if (!noRetry) {
+    try {
+      console.log(`EC2 id ${await SpotNotifier.instanceId()}`);
+      SpotNotifier.on('termination', () => shutdown(opts));
+      SpotNotifier.start();
+    } catch (err) {
+      console.log('SpotNotifier can not be started.');
+    }
   }
 
   if (parseInt(idleTimeout) !== 0) {
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -234,12 +234,34 @@ const runLocal = async (opts) => {
     } else if (log && log.status === 'job_ended') {
       const { job } = log;
 
+      const waitCompletedPipelineJobs = () => {
+        return new Promise((resolve, reject) => {
+          try {
+            if (RUNNER_JOBS_RUNNING.length === 1) {
+              resolve([RUNNER_JOBS_RUNNING[0].id]);
+              return;
+            }
+
+            const watcher = setInterval(async () => {
+              const jobs = (
+                await cml.pipelineJobs({ jobs: RUNNER_JOBS_RUNNING })
+              )
+                .filter((job) => job.status === 'completed')
+                .map((job) => job.id);
+
+              if (jobs.length) {
+                resolve(jobs);
+                clearInterval(watcher);
+              }
+            }, 5 * 1000);
+          } catch (err) {
+            reject(err);
+          }
+        });
+      };
+
       if (!RUNNER_SHUTTING_DOWN) {
-        const jobs = job
-          ? [job]
-          : (await cml.pipelineJobs({ jobs: RUNNER_JOBS_RUNNING }))
-              .filter((job) => job.status === 'completed')
-              .map((job) => job.id);
+        const jobs = job ? [job] : await waitCompletedPipelineJobs();
 
         RUNNER_JOBS_RUNNING = RUNNER_JOBS_RUNNING.filter(
           (job) => !jobs.includes(job.id)
@@ -253,7 +275,7 @@ const runLocal = async (opts) => {
   proc.on('uncaughtException', () => shutdown(opts));
   proc.on('disconnect', () => shutdown(opts));
 
-  if (cml.driver === 'github' && parseInt(idleTimeout) !== 0) {
+  if (parseInt(idleTimeout) !== 0) {
     const watcher = setInterval(() => {
       RUNNER_TIMEOUT_TIMER > idleTimeout &&
         shutdown(opts) &&
@@ -263,7 +285,7 @@ const runLocal = async (opts) => {
     }, 1000);
   }
 
-  if (cml.driver === 'github' && !noRetry) {
+  if (!noRetry && cml.driver === 'github') {
     const watcher = setInterval(async () => {
       RUNNER_JOBS_RUNNING.forEach((job) => {
         if (
@@ -272,11 +294,6 @@ const runLocal = async (opts) => {
         )
           shutdown(opts) && clearInterval(watcher);
       });
-
-      const runner = await cml.runnerByName({ name: name });
-      if (runner && runner.status && runner.status.toLowerCase() === 'idle') {
-        RUNNER_JOBS_RUNNING = [];
-      }
     }, 60 * 1000);
   }
 
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -5,6 +5,7 @@ const { homedir } = require('os');
 
 const fs = require('fs').promises;
 const yargs = require('yargs');
+const { SpotNotifier } = require('ec2-spot-notification');
 
 const { exec, randid, sleep } = require('../src/utils');
 const tf = require('../src/terraform');
@@ -214,7 +215,7 @@ const runCloud = async (opts) => {
 
 const runLocal = async (opts) => {
   console.log(`Launching ${cml.driver} runner`);
-  const { workdir, name, labels, single, idleTimeout, noRetry } = opts;
+  const { workdir, name, labels, single, idleTimeout, noRetry, cloud } = opts;
 
   const proc = await cml.startRunner({
     workdir,
@@ -223,8 +224,6 @@ const runLocal = async (opts) => {
     single,
     idleTimeout
   });
-
-  console.log(proc);
 
   const dataHandler = async (data) => {
     const log = await cml.parseRunnerLog({ data });
@@ -277,6 +276,12 @@ const runLocal = async (opts) => {
   proc.on('uncaughtException', () => shutdown(opts));
   proc.on('disconnect', () => shutdown(opts));
   proc.on('exit', () => shutdown(opts));
+
+  if (cloud === 'aws') {
+    await SpotNotifier.instanceId();
+    SpotNotifier.on('termination', () => shutdown(opts));
+    SpotNotifier.start();
+  }
 
   if (parseInt(idleTimeout) !== 0) {
     const watcher = setInterval(() => {
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -294,7 +294,7 @@ const runLocal = async (opts) => {
   }
 
   if (!noRetry && cml.driver === 'github') {
-    const watcher = setInterval(async () => {
+    const watcher = setInterval(() => {
       RUNNER_JOBS_RUNNING.forEach((job) => {
         if (
           new Date().getTime() - new Date(job.date).getTime() >
--- a/bin/cml-runner.js
+++ b/bin/cml-runner.js
@@ -224,6 +224,8 @@ const runLocal = async (opts) => {
     idleTimeout
   });
 
+  console.log(proc);
+
   const dataHandler = async (data) => {
     const log = await cml.parseRunnerLog({ data });
     log && console.log(JSON.stringify(log));
@@ -274,6 +276,7 @@ const runLocal = async (opts) => {
   proc.stdout.on('data', dataHandler);
   proc.on('uncaughtException', () => shutdown(opts));
   proc.on('disconnect', () => shutdown(opts));
+  proc.on('exit', () => shutdown(opts));
 
   if (parseInt(idleTimeout) !== 0) {
     const watcher = setInterval(() => {
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -241,7 +241,8 @@ class Github {
       name,
       labels: labels.map(({ name }) => name),
       online: status === 'online',
-      busy
+      busy,
+      status
     }));
   }
 
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -258,8 +258,7 @@ class Github {
       name,
       labels: labels.map(({ name }) => name),
       online: status === 'online',
-      busy,
-      status
+      busy
     }));
   }
 
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -176,15 +176,14 @@ class Gitlab {
     const endpoint = `/runners?per_page=100`;
     const runners = await this.request({ endpoint, method: 'GET' });
     return await Promise.all(
-      runners.map(async ({ id, description, active, online, status }) => ({
+      runners.map(async ({ id, description, active, online }) => ({
         id,
         name: description,
         labels: (
           await this.request({ endpoint: `/runners/${id}`, method: 'GET' })
         ).tag_list,
         online,
-        busy: active && online,
-        status
+        busy: active && online
       }))
     );
   }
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: bin/cml-runner.js [hunk 1]
```diff
@@ -5,6 +5,7 @@ const { homedir } = require('os');
 
 const fs = require('fs').promises;
 const yargs = require('yargs');
+const { SpotNotifier } = require('ec2-spot-notification');
```
0x2b3bfa0, line 8: Can we separate this functionality and move it to the AMI on the terraform-provider-iterative repository? While this is a must have, keeping it as part of the CML code would introduce another unrelated responsibility to this codebase.

0x2b3bfa0: We can just `curl` the `169.254.169.254` endpoint in a loop and run `systemctl stop cml` once we get a termination notice, all from the instance script.

DavidGOrtega: Lets open another ticket and use for now the already done work that also works.

DavidGOrtega: Also, we might just want this behaviour with the runner, so I guess it should go in the TPI runner resource at most

DavidGOrtega: > Can we separate this functionality and move it to the AMI and fallback on the terraform-provider-iterative repository? While this is a must have, keeping it as part of the CML code would introduce another unrelated responsibility to this codebase.

Actually I disagree, Its a part of the logic that we are setting to handle long job runners. The same that we do with the 72 hours in example. And ensures the integrty and responsability of the runner

0x2b3bfa0: > _Also, we might just want this behaviour with the runner, so I guess it should go in the TPI runner resource at most_

Agreed!

DavidGOrtega: Putting it outside makes more difficult to follow up the logic that ensures this feature of the runner. It MUST be definitely inside the runner

0x2b3bfa0: > _Actually I disagree, Its a part of the logic that we are setting to handle long job runners. The same that we do with the 72 hours in example. And ensures the integrty and responsability of the runner_

I also disagree in turn. 😈  This is a cloud-specific termination detection mechanism and, no matter what its purpose is, should ideally be handled at the `resource_cml_runner` level at most. This project should not care about clouds but only about SCM platforms, and use `SIGTERM` as a common signal for every termination scenario.

DavidGOrtega: > should ideally be handled at the resource_cml_runner level at most. This project should not care about clouds but only about SCM platforms, and use SIGTERM as a common signal for every termination scenario.

Why? why SCMs and not clouds? Who puts those constrains? Putting it outside its not a good idea. Indeed Im missing someting... whatch next commit

0x2b3bfa0: > _Putting it outside makes more difficult to follow up the logic that ensures this feature of the runner. It MUST be definitely inside the runner_

More than a "feature" of the runner, I would consider it a cloud-vendor-specific patch to notify processes on spot termination with the same mechanism used for instance termination.

DavidGOrtega: I think that the latest commit closes the discussion

0x2b3bfa0: You're making the AWS spot instance termination detection logic optional, but...

1. How are we detecting the termination of Azure spot and Google Compute preemptive instances?
2. Do you still think that it's a good idea to introduce code specific to cloud vendors in the CML repository?

DavidGOrtega: > Do you still think that it's a good idea to introduce code specific to cloud vendors in the CML repository?

yes, I would like to have all the base code together. If it were for me the TPI would be done in js... actually... I have a proposal 😉 

> How are we detecting the termination of Azure spot and Google Compute preemptive instances?

It does not exists. SIGTERM should suffice. Actually in AWS is intermittent. Seems that SIGTERM does not always work. ideally we should not need it.

DavidGOrtega: Anyway if the code works correctly and the direction is not clear is better to release and USE already developed and tested code

0x2b3bfa0: > _yes, I would like to have all the base code together. If it were for me the TPI would be done in js... actually... I have a proposal_

I would too, if Hashicorp had chosen to implement Terrraform in JavaScript. Given that it's not the case, I'd prefer to separate the responsibilities where they make most sense, keeping all the cloud-specific logic in the provider and all the SCM-specific logic in CML, as we do now.

> _It does not exists. SIGTERM should suffice. Actually in AWS is intermittent. Seems that SIGTERM does not always work. ideally we should not need it._

Then, shouldn't this be a patch on the provider so we convert AWS termination notifications to a normal `SIGTERM` signal? If you can't tell the difference between spot termination and instance terminaltion in other vendors, why do it differently on this one?

src/drivers/github.js
```diff
@@ -258,7 +258,8 @@ class Github {
       name,
       labels: labels.map(({ name }) => name),
       online: status === 'online',
-      busy
+      busy,
+      status
```
0x2b3bfa0, line 262: Why is this needed?

src/drivers/gitlab.js
```diff
@@ -176,14 +176,15 @@ class Gitlab {
...
         labels: (
           await this.request({ endpoint: `/runners/${id}`, method: 'GET' })
         ).tag_list,
         online,
-        busy: active && online
+        busy: active && online,
+        status
       }))
     );
   }
```
0x2b3bfa0, line 190: Why is this needed? Same as https://github.com/iterative/cml/pull/653/files#r673025416

bin/cml-runner.js
DavidGOrtega: > I would too, if Hashicorp had chosen to implement Terrraform in JavaScript. Given that it's not the case,

Im thinking to do all in JS executed with V8. At the very end we are using the vendors SDK that already exists in JS

> Then, shouldn't this be a patch on the provider so we convert AWS termination notifications to a normal SIGTERM signal? If you can't tell the difference between spot termination and instance terminaltion in other vendors, why do it differently on this one?

Reasons: 

 - More complex in the TPI
 - More obscure having another service or whatever...
 - This runner could be launched manually inside an AWS instance and still works on its own

src/drivers/github.js
DavidGOrtega: We were doing it before, we need to check the status of completion

src/drivers/gitlab.js
DavidGOrtega: To be consistent with something that we were doing

0x2b3bfa0: Let's move the conversation to https://github.com/iterative/cml/pull/653/files#r673025416

src/drivers/github.js
0x2b3bfa0: Where are we using the value of this property in the code? 🤔

0x2b3bfa0: And, what's the relationship between the `status` returned by this function and the completion of a job?

DavidGOrtega:
```js
.filter((job) => job.status === 'completed')
```

DavidGOrtega: Its true. I added it to check the runner status with the first solution

0x2b3bfa0: That `filter` is taking the output from `cml.pipelineJobs` and not `cml.runners` unless I missed something. 🤔

DavidGOrtega: Anyway is a nice field

DavidGOrtega: > That filter is taking the output from cml.pipelineJobs and not cml.runners unless I missed something. 🤔

Yes,I remembered the first approach that were runners indeed

0x2b3bfa0: I'm no expert in measuring the niceness of fields, but it has been superseded by `online` and `busy` fields, that offer the same values across all the supported platforms.

src/drivers/gitlab.js
```diff
@@ -176,14 +176,15 @@ class Gitlab {
...
+      runners.map(async ({ id, description, active, online, status }) => ({
         id,
         name: description,
         labels: (
           await this.request({ endpoint: `/runners/${id}`, method: 'GET' })
         ).tag_list,
         online,
-        busy: active && online
+        busy: active && online,
+        status
```
0x2b3bfa0, line 187:
```suggestion
        busy: active && online
```

src/drivers/github.js
DavidGOrtega: It might not be enough actually. And i would say that I would deprecate online and busy in favour of status which is richer and something that we are exposing for the viewer

0x2b3bfa0: The `status` field doesn't offer the same values across platforms, while `online` and `busy` do. Nevertheless, we should consider this separately; the `status` field is not necessary for this pull request.

src/drivers/gitlab.js
```diff
@@ -176,14 +176,15 @@ class Gitlab {
     const endpoint = `/runners?per_page=100`;
     const runners = await this.request({ endpoint, method: 'GET' });
     return await Promise.all(
-      runners.map(async ({ id, name, description, active, online }) => ({
+      runners.map(async ({ id, description, active, online, status }) => ({
```
0x2b3bfa0, line 179:
```suggestion
      runners.map(async ({ id, description, active, online }) => ({
```
See [this conversation](https://github.com/iterative/cml/pull/653/files/fe8831ad8674554d0081ebfba0d83e4f24bf7f75#r673025416) for context.

src/drivers/github.js
0x2b3bfa0:
```suggestion
```

bin/cml-runner.js
```diff
@@ -234,12 +235,34 @@ const runLocal = async (opts) => {
...
+        return new Promise((resolve, reject) => {
+          try {
+            if (RUNNER_JOBS_RUNNING.length === 1) {
+              resolve([RUNNER_JOBS_RUNNING[0].id]);
+              return;
+            }
+
+            const watcher = setInterval(async () => {
+              const jobs = (
+                await cml.pipelineJobs({ jobs: RUNNER_JOBS_RUNNING })
```
0x2b3bfa0, line 248: The GitLab driver does not have `pipelineJobs` 🤔

DavidGOrtega: Its not needed for it

0x2b3bfa0: After https://github.com/iterative/cml/pull/653/commits/082fcba905e2708de9d013af9a222e6fc2a98438 it won't throw a TypeError when calling an undefined function. Still, I suspect that it's going to throw the `not implemented` error you introduce.

Should we call `waitCompletedPipelineJobs` only if `driver === 'github'` or something?

DavidGOrtega: No, actually that will never be called in Gitlab. We maintain it to give coherence and to establish the interface. n Gitalb that will never ever be called

DavidGOrtega: Let me clarify. In Gitlab the logs comes with the jobID thing that does not happen in Github and we have to gather the jobs to check the completion of the id we knew

0x2b3bfa0: > _No, actually that will never be called in Gitlab_ <sup>[citation needed]</sup>

Is there any guard code to prevent that? 🤔

DavidGOrtega: > Is there any guard code to prevent that? 🤔

the existance of the job.id

DavidGOrtega:
```js
const jobs = job ? [job] : await waitCompletedPipelineJobs();
```

DavidGOrtega: As  say in GL that id is known always in the log

0x2b3bfa0: 👌🏼  Thanks!

bin/cml-runner.js
0x2b3bfa0: > _Im thinking to do all in JS executed with V8. At the very end we are using the vendors SDK that already exists in JS_

Not sure about the specific advantages of V8 over any other engine for this use case, but anyhow. This is a completely separate discussion topic.

> * More complex in the TPI

Not _that_ complex.

> * More obscure having another service or whatever...

Not as obscure as moving cloud-specific code to the CML repository instead of abstracting over signals, in my opinion.

> * This runner could be launched manually inside an AWS instance and still works on its own

Do we _really_ want to support this use case? What's the advantage of supporting it if we don't do the same for the other providers?

DavidGOrtega: All advantages despite that you disagree.

DavidGOrtega: > Not as obscure as moving cloud-specific code to the CML repository

CML repository is a runner also that can check whatever is needed to ensure the feature it offers. Transparent spot instances

DavidGOrtega: > Not that complex.

We have to check also the noRetry... pass it.. create another service or whatever, business logic broken in multiple parts ... Ugly! 
I can't believe that you can not see the benefit

DavidGOrtega: > Not sure about the specific advantages of V8 over any other engine for this use case, but anyhow. This is a completely separate discussion topic.

Im telling you a solution to avoid Golang source code that Im planning to also unify the base source code that we were discussing

DavidGOrtega:
@0x2b3bfa0 are you approving it? We have several p0s that are fixed here. I want to release it ASAP

0x2b3bfa0:
I'm hesitant to approve the inclusion of SpotNotifier in the CML code base, especially without a follow-up technical debt / discussion issue.

I'm going to run tests on Google Cloud to confirm that #661 gets fixed and check everything for approval, but I'd prefer to discuss this in the weekly meeting so we can take better decisions.

DavidGOrtega:
> I'm hesitant to approve the inclusion of SpotNotifier in the CML code base, especially without a follow-up technical debt / discussion issue.

Open a ticket to follow up. I do agree with it and I wont agree the opposite, so... To me there is not technical debt here

> I'm going to run tests on Google Cloud to confirm that #661 gets fixed and check everything for approval, but I'd prefer to discuss this in the weekly meeting so we can take better decisions.

I have tested it. The machines are disposed

DavidGOrtega:
opens #663

0x2b3bfa0 (approved):
Looks good to me, except for the inclusion of SpotNotifier. Approving just because of https://github.com/iterative/cml/pull/653#issuecomment-883373907 to avoid blocking priority 0 fixes.

ivyleavedtoadflax:
@DavidGOrtega @0x2b3bfa0 thanks so much for this!

lemontheme:
Reporter of #661 checking in. Indeed, thanks for jumping on this!

Unnnnnfortunately, I'm afraid that I'm still seeing the same behavior as before. (I'm assuming that the iterative/setup-cml@v1 action uses the latest cml release.) In 5 out of 5 separate runs, my Google CE instances stayed alive rather than being shutdown.

Seeing as your tests passed though, maybe it's user error on my end?

DavidGOrtega:
@lemontheme I would need to see your workflow. Tests with TPI indicates that the instances are disposed after the expected time.

lemontheme:
@DavidGOrtega sorry for the late reply. I tested with the same workflow as in #661

0x2b3bfa0:
@lemontheme, can you please open a new issue?

PR #731, opened 2021-09-22
Fix runner status log reporting

0x2b3bfa0:
Missing the <kbd>blunder</kbd> label

### DIFF
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -229,7 +229,7 @@ const runLocal = async (opts) => {
 
   const dataHandler = async (data) => {
     const log = await cml.parseRunnerLog({ data });
-    log && winston.info("runner log", log);
+    log && winston.info('runner log', log);
 
     if (log && log.status === 'job_started') {
       RUNNER_JOBS_RUNNING.push({ id: log.job, date: log.date });
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -229,7 +229,7 @@ const runLocal = async (opts) => {
 
   const dataHandler = async (data) => {
     const log = await cml.parseRunnerLog({ data });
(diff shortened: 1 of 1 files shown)


bin/cml/runner.js
```diff
@@ -229,7 +229,7 @@ const runLocal = async (opts) => {
 
   const dataHandler = async (data) => {
     const log = await cml.parseRunnerLog({ data });
-    log && winston.debug(JSON.stringify(log));
+    log && winston.info("runner log", log);
```
0x2b3bfa0, line 232:
```suggestion
    log && winston.info('runner log', log);
```

DavidGOrtega (approved):
👍 lgtm

0x2b3bfa0:
### Explanation

Winston, the logging library introduced with #703, only outputs plain text on interactive environments. When the standard output is not a TTY, it will output plain [JSONL](https://jsonlines.org/) logs instead.

https://github.com/iterative/cml/blob/7d664aec00023e8fba48e660f2f8e184477d8cdd/bin/cml.js#L12-L17

The provided `winston.debug` call was outputting the JSON object as an escaped JSON string embedded in another object, thus causing the Terraform provider to time out waiting for the readiness signal, which was unable to parse.

***

Not a proper report, but enough to cover the causes. I still have lots of things to do today. 🏃🏼 😅

casperdcl:
yup

xkcd#844

0x2b3bfa0:
_Cough, cough._ Actually, that wasn't supposed to be `winston.debug` but a plain `console.log` in the first place; see https://github.com/iterative/cml/pull/703/commits/8de7c23b78c654d97e094bc838fce630b11257c6

casperdcl:
:airplane:

PR #739, opened 2021-09-22
Manage all the environment variables through yargs

0x2b3bfa0:
Would close #738, depends on #712

PR #712, opened 2021-09-09
Convert yargs builder functions to options objects and add per–command environment variables

0x2b3bfa0:
Fixes https://github.com/iterative/cml/issues/708 along with some minor creative bugs we've been overlooking for months. 🐞

**Contains cascaded https://github.com/iterative/cml/pull/739**

PR #742, opened 2021-09-23
Patch runner name while finding a better solution

0x2b3bfa0:


DavidGOrtega:
@0x2b3bfa0 do you think this will just suffice? Should we not even add here the CML_ prefix and warn them to replace it one go?

casperdcl (changes requested):
I would have a more helpful warning. It should state what to do.

Either:

- use `CML_RUNNER_NAME`, or
- use `--name`

0x2b3bfa0:
If we merge #712, there is a nicer fix with #739

0x2b3bfa0:
> _I would have a more helpful warning._

Absolutely! :100:

0x2b3bfa0:
Closing in favor of https://github.com/iterative/cml/pull/739

PR #765, opened 2021-10-12
Rely on proxy-agent for detecting environment variables

### LINKED ISSUE #760: Environment variable "no_proxy" seems to be unsupported
I've been trying to play around with self hosted runners. 

I'm sitting behind a corporate firewall using Github Enterprise, which is also inside the firewall.

I'm using iterativeai/cml:0-dvc2-base1-gpu docker container on a Linux Ubuntu 18.04 machine.
I am fairly confident, my container is set up properly with certificates and proxy settings. For instance can I clone my Github Enterprise repo with no problems. I am also able without any problems to connect an original github runner to my Enterprise repo.

However when I use cml runner, I seem to fail in both of two options:
a) I pass my proxy settings to the docker container, which makes it impossible to reach my Github Enterprise repository (because the proxy is not able to resolve the address).
b) I leave out the proxy settings, so the CML runner can find the Github Enterprise but when it next tries to load the latest Github runner version from github.com, it fails because it doesn't get past the proxy.

It seems to me, that the problem origins in the usage of node-fetch.js, which in turn also only supports either using http/s_proxy or not. But there is no gate evaluating the requested URL and changing the path either through the proxy or not. I'm really no expert in this as you might have guessed after the first few words. So let me try to phrase it in my simple words:

Can you implement cml runner to "respect" the no_proxy setting? Or help me in any other way?

0x2b3bfa0:
> _If no URI is provided, then [proxy-from-env](https://www.npmjs.com/package/proxy-from-env) is used to get the URI from `$HTTP_PROXY`, `$HTTPS_PROXY` and `$NO_PROXY` among others._ **(Excerpt from [`proxy-agent` documentation](https://www.npmjs.com/package/proxy-agent))**

This pull request will fix #760 by honoring the [`no_proxy`](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy) environment variable and add compatibility for proxy types other than `https` as listed on [this table](https://www.npmjs.com/package/proxy-agent).


0x2b3bfa0:
Note for testers (hello, testers): you can use [`mitmproxy`](https://mitmproxy.org/) to verify whether the `no_proxy` environment variable is being honored or not.

Spoiler: it works as expected.

PR #797, opened 2021-11-01
Add -no-color flag to tf apply

dacbd:
Sorry for the previous one :octocat:

### DIFF
--- a/src/terraform.js
+++ b/src/terraform.js
@@ -37,7 +37,12 @@ const init = async (opts = {}) => {
 
 const apply = async (opts = {}) => {
   const { dir = './' } = opts;
-  return await exec(`terraform -chdir='${dir}' apply -auto-approve -no-color`);
+  const { CI } = process.env;
+  return await exec(
+    `terraform -chdir='${dir}' apply -auto-approve ${
+      CI === 'true' ? '-no-color' : ''
+    }`
+  );
 };
 
 const destroy = async (opts = {}) => {
--- a/src/terraform.js
+++ b/src/terraform.js
@@ -37,7 +37,7 @@ const init = async (opts = {}) => {
 
 const apply = async (opts = {}) => {
   const { dir = './' } = opts;
-  return await exec(`terraform -chdir='${dir}' apply -auto-approve`);
+  return await exec(`terraform -chdir='${dir}' apply -auto-approve -no-color`);
 };
 
 const destroy = async (opts = {}) => {


casperdcl (commented):
does colour really not work on any CI provider (before this PR)?

dacbd:
I actually forgot the specific instance I encountered the polluted logs and can't seem to find it again... so no worries.

dacbd:
image

dacbd:
With patch:
image

0x2b3bfa0:
Note: we output JSONL logs when `cml` does not run in a TTY so they are easier to parse, but you can always use `jq` to see the human–readable version, even with colors.

```console
jq --slurp --raw-output '.[].message' < logs.jsonl
```

dacbd:
@0x2b3bfa0 I maybe I should look closely at how the logging is being done, there could be a better solution.

These screenshots wherefrom the GitHub actions interface. Ideally, the output should be as clear as possible. The required information to find the problem was presented but we should not have to go through extra steps to be able to quickly read them?

I feel like there is some inconsistency in their presentation. It's hard for me to go back and look at my failed logs for it because introduced my `-no-color` patch in with my [setup-cml](https://github.com/iterative/setup-cml/pull/51/commits/e3d8f8499dc2f44fc1ae8e5f84669525cb6c0ef5) fork.

With the json format `info` level logging prints the `message` without color, but for `error` level captures the color in `message`. 
I think this should not be the case, right?

Below was the quickest way I could demonstrate the problem:
`$ bash -e ./test.sh 1> /dev/null`
then stop with `^C`

0x2b3bfa0:
I wonder if we should be “less smart” about output formats and always output human–readable text unless the `--json` flag is specified.

https://github.com/iterative/cml/blob/68ddbe4bfa329e5a8b90caa3c7f9aa8a8acda109/bin/cml.js#L12

0x2b3bfa0 (approved):
🔔 @iterative/cml, voting to merge this if it's helpful for @dacbd, but we should probably look into outputting human–readable logs by default and creating a command–line option to explicitly enable JSONL logs from TPI.

dacbd:
Could be, I think the JSON format is readable enough for most.

Now I am more curious to find out why the like I said before why the info level logging filters out colors in `message` but error level does not.

casperdcl:
does it really make sense to strip colour just from `apply`? Surely it should be for all commands... but then at that stage it would probably be a better idea to fix the real issue (https://github.com/iterative/cml/pull/797#issuecomment-962750850 or https://github.com/iterative/cml/pull/797#issuecomment-962718030)

dacbd:
@casperdcl I agree, I'm good with closing this in favor of a new issue.

dacbd:
Closing in favor of: https://github.com/iterative/cml/issues/806

PR #813, opened 2021-11-17
Improve BitBucket unauthorized error message

duijf:
Was playing around with CML for some integration work with Studio when I encountered this papercut.

On 401 unauthorized responses, the BitBucket API does not always return a JSON body. That was causing the error handling code that's being changed here to throw an error itself, which led to errors like:

```
error: invalid json response body at [url] reason: Unexpected end of JSON input {"type":"invalid-json"}
```

Which was a bit confusing to me. This commit changes the error handling code to return a generic error message instead. So with an incorrect BitBucket token, CML now prints:

```
error: Unauthorized
```

Which should be a little more helpful to the user.

NB: My JS is not the best ever, so if something looks weird / unidiomatic, it probably is. Please let me know if you'd like to see any changes :)

### DIFF
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -253,10 +253,14 @@ class BitbucketCloud {
     });
 
     if (response.status > 300) {
-      const {
-        error: { message }
-      } = await response.json();
-      throw new Error(`${response.statusText} ${message}`);
+      try {
+        const {
+          error: { message }
+        } = await response.json();
+        throw new Error(`${response.statusText} ${message}`);
+      } catch (err) {
+        throw new Error(`${response.statusText}`);
+      }
     }
 
     return await response.json();
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -257,9 +257,9 @@ class BitbucketCloud {
         const {
           error: { message }
         } = await response.json();
-        throw new Error(`${response.statusText} ${message}`);
+        throw new Error(message);
       } catch (err) {
-        throw new Error(`${response.statusText}`);
+        throw new Error(`${response.statusText} ${err.message}`);
       }
     }
 


src/drivers/bitbucket_cloud.js
```diff
@@ -253,10 +253,14 @@ class BitbucketCloud {
...
-      } = await response.json();
-      throw new Error(`${response.statusText} ${message}`);
+      try {
+        const {
+          error: { message }
+        } = await response.json();
+        throw new Error(`${response.statusText} ${message}`);
+      } catch (err) {
+        throw new Error(`${response.statusText}`);
+      }
```
0x2b3bfa0, line 263:
```suggestion
      try {
        const {
          error: { message }
        } = await response.json();
        throw new Error(message);
      } catch (err) {
        throw new Error(`${response.statusText} ${err.message}`);
      }
```

I was wondering if we could take advantage of the `catch` block to avoid repeating `response.statusText` in the error message. 🤔

0x2b3bfa0 (approved):
:shipit:

src/drivers/bitbucket_cloud.js
0x2b3bfa0: It would have worked on languages that use `return` and error checking instead of `throw` and error handling. 🙊

duijf:
GitLab CI failed with an error from Terraform. I don't have enough context to easily see what's happening there, but it looks like a timeout that's unrelated to the changes here.

0x2b3bfa0:
* https://github.com/iterative/viewer-itops/pull/117

0x2b3bfa0:
> _GitLab CI failed with an error from Terraform. I don't have enough context to easily see what's happening there, but it looks like a timeout that's unrelated to the changes here._

This happens because the spot instance request didn't get fulfilled before the 10 minute deadline. The error is because we lack enough permissions to remove the request, but it would have failed anyway.

DavidGOrtega:
Thanks for the contribution ❤️ @duijf  now we feel obligated to break Viewer... somehow 😁

PR #816, opened 2021-11-19
Better error reporting/debugging for BitBucket

duijf:
Continued from https://github.com/iterative/cml/pull/813, this PR improves the logging related to BitBucket API calls.

## Summary

This addresses two papercuts / UX problems:

 - Prior to this PR, there was no debug logging of the actual HTTP requests that are being made to BitBucket.
 - BitBucket sometimes does not send error responses in the documented format. This would cause subpar error messages.

## Background

I tried to make CML post BitBucket comments on a commit. The API requests were failing because I did not have the "Pull requests links" application installed (this requirement is documented [here](https://cml.dev/doc/ref/send-comment#bitbucket))

This didn't work. CML gave me the following output:

```
error: Bad Request undefined
```

With `--log debug`, I also didn't get enough information about what was happening to debug the problem. I had to add the debug logging to CML directly.

The error response in this case was slightly different than what BitBucket sends normally (and what their documentation says it they send):

 - The API response was `{"error": "Invalid or unknown installation"}`.
 - The code assumed that BitBucket was always returning responses in the format `{"error": {"message": "Some message"}}`

That was causing the `undefined` to be printed.

The lack of context made things take a bit longer to debug than I wanted. I would have expected a request log when I passed `--log debug`, but it turns out we didn't have any debug logging for these requests yet. That meant I had to add the debug logging myself to get the full API response from BB.

## Changes

 - Handle the undocumented error response format in addition to the format that we handled already. Fall back to an empty string, so we don't print `undefined`.
 - Add debug logging for requests / responses so users have more information available to inspect what's happening when things go wrong.

Let me know if we want to add tests for this or want to reorganize the code somewhat :)

### DIFF
--- a/src/drivers/bitbucket_cloud.js
+++ b/src/drivers/bitbucket_cloud.js
@@ -1,4 +1,5 @@
 const fetch = require('node-fetch');
+const winston = require('winston');
 const { URL } = require('url');
 const ProxyAgent = require('proxy-agent');
 
@@ -245,7 +246,10 @@ class BitbucketCloud {
       Authorization: 'Basic ' + `${token}`
     };
 
-    const response = await fetch(url || `${api}${endpoint}`, {
+    const requestUrl = url || `${api}${endpoint}`;
+    winston.debug(`${method} ${requestUrl}`);
+
+    const response = await fetch(requestUrl, {
       method,
       headers,
       body,
@@ -254,10 +258,12 @@ class BitbucketCloud {
 
     if (response.status > 300) {
       try {
-        const {
-          error: { message }
-        } = await response.json();
-        throw new Error(message);
+        const json = await response.json();
+        winston.debug(json);
+        // Attempt to get additional context. We have observed two different error schemas
+        // from BitBucket API responses: `{"error": {"message": "Error message"}}` and
+        // `{"error": "Error message"}`.
+        throw new Error(json.error.message || json.error);
       } catch (err) {
         throw new Error(`${response.statusText} ${err.message}`);
       }


src/drivers/bitbucket_cloud.js
```diff
@@ -254,9 +258,12 @@ class BitbucketCloud {
...
-        const {
-          error: { message }
-        } = await response.json();
+        const json = await response.json();
+        winston.debug(json);
+        // We have observed BitBucket sometimes sends { "error": { "message": "Some message" } }
+        // and sometimes { "error": "Some message" }. Attempt to get additional context. Fall back
+        // to the empty string to ensure we don't print `undefined` to the user.
+        const message = json.error.message || json.error || '';
         throw new Error(message);
```
0x2b3bfa0, line 267:
```suggestion
        throw new Error(json.error.message || json.error || '');
```
Would it be better to merge theste two lines into a single one?

duijf: Thanks for the suggestion! Agreed & updated :)

src/drivers/bitbucket_cloud.js
```diff
@@ -254,10 +258,13 @@ class BitbucketCloud {
...
-          error: { message }
-        } = await response.json();
-        throw new Error(message);
+        const json = await response.json();
+        winston.debug(json);
+        // Attempt to get additional context. We have observed two different error schemas
+        // from BitBucket API responses: `{"error": {"message": "Error message"}}` and
+        // `{"error": "Error message"}`. Fall back to the empty string to ensure we don't
+        // print `undefined` to the user.
+        throw new Error(json.error.message || json.error || '');
```
0x2b3bfa0, line 267:
```suggestion
        throw new Error(json.error.message || json.error);
```

It looks like `new Error(undefined).message === ""` and there would be no need of providing an empty string fallback.

0x2b3bfa0 (approved):

src/drivers/bitbucket_cloud.js
duijf: Ah you're right!

Retraced my steps to see how I managed to get `undefined` in the error message. Seems like I was running `0.9.0` instead of `master` when I got this output. So that bit was already fixed by https://github.com/iterative/cml/commit/44806c0d583fe359fdf34389332d802975f6f501 :smile:

PR #818, opened 2021-11-20
`cml runner` - Cleaner terraform JSON logs

### LINKED ISSUE #806: `cml runner` inconsistent logging/capture style of terraform output
For Context: originally discussed on PR #797 

> With the json format `info` level logging prints the `message` without color, but for `error` level captures the color in `message`. 
> I think this should not be the case, right?

> Below was the quickest way I could demonstrate the problem:
> `$ bash -e ./test.sh 1> /dev/null`
> then stop with `^C`

_Originally posted by @dacbd in https://github.com/iterative/cml/issues/797#issuecomment-962655261_

I'll spend some time taking another look at the logging and how the stdout/err are captured from the invoked TF commands.

dacbd:
I tried a couple of iterations to solve this and settled with this.

I named the regex groups to hopefully make reading the regex easier for others in the future (I know I hate figuring out what they do 🔮 ).

`\x1b[31;4mtext`
- `\x1b` or `\u001b` escape char
- escape char + `[` is the "Control Sequence Induce"
- `31;4` followed by `m` forms the "Select Graphic Rendition" the former being parameter bytes and the latter a termination byte

Closes #806

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,7 +11,7 @@ module.exports = {
     SharedArrayBuffer: 'readonly'
   },
   parserOptions: {
-    ecmaVersion: 2018
+    ecmaVersion: 2020
   },
   ignorePatterns: ['assets/', 'dist/', 'node_modules/'],
   rules: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,7 +11,7 @@ module.exports = {
     SharedArrayBuffer: 'readonly'
   },
   parserOptions: {
-    ecmaVersion: 2020
+    ecmaVersion: 2018
   },
   ignorePatterns: ['assets/', 'dist/', 'node_modules/'],
   rules: {
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "pseudoexec": "^0.2.0",
     "semver": "^7.3.5",
     "simple-git": "^2.46.0",
-    "strip-ansi": "^7.0.1",
+    "strip-ansi": "^6.0.1",
     "strip-url-auth": "^1.0.1",
     "tar": "^6.1.11",
     "tempy": "^0.6.0",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "pseudoexec": "^0.2.0",
     "semver": "^7.3.5",
     "simple-git": "^2.46.0",
+    "strip-ansi": "^7.0.1",
     "strip-url-auth": "^1.0.1",
     "tar": "^6.1.11",
     "tempy": "^0.6.0",
--- a/src/utils.js
+++ b/src/utils.js
@@ -13,7 +13,7 @@ const exec = async (command) => {
       (error, stdout, stderr) => {
         if (!process.stdout.isTTY) {
           const controlSequence =
-            /(?<excape_char>\\u001b|\u001b)(?<csi_term_char>\[)(?<parameters>(\d*;?)+)(?<term_byte>m)/g; // eslint-disable-line no-control-regex
+            /(?<csi_char>(\\u001b|\u001b)\[)(?<sgr_bytes>(\d*;?)+)(?<sgr_term>m)/g; // eslint-disable-line no-control-regex
           stdout = stdout.replace(controlSequence, '');
           stderr = stderr.replace(controlSequence, '');
         }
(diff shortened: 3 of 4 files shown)


src/utils.js
```diff
@@ -11,6 +11,12 @@ const exec = async (command) => {
       command,
       { ...process.env },
       (error, stdout, stderr) => {
+        if (!process.stdout.isTTY) {
+          const controlSequence =
+            /(?<csi_char>(\\u001b|\u001b)\[)(?<sgr_bytes>(\d*;?)+)(?<sgr_term>m)/g; // eslint-disable-line no-control-regex
```
0x2b3bfa0, line 16:
```suggestion
            /(?<csi_char>\u001b\[)(?<sgr_bytes>(\d*;?)+)(?<sgr_term>m)/g; // eslint-disable-line no-control-regex
```

Shouldn't we only match the `␛` character?

0x2b3bfa0:
I wonder if we could use [`strip-ansi`](https://github.com/chalk/strip-ansi) instead. It's backed by [`ansi-regex`](https://github.com/chalk/ansi-regex) and its “exceedingly compliant” test suite.

src/utils.js
dacbd: It never happened in my testing but I threw in the literal form just incase in came up. It never did with what I tested to taking it out works for me.

dacbd:
> I wonder if we could use [`strip-ansi`](https://github.com/chalk/strip-ansi) instead. It's backed by [`ansi-regex`](https://github.com/chalk/ansi-regex) and its “exceedingly compliant” test suite.

That is certainly a more robust solution. I wanted to avoid much bloat and didn't want to introduce any external dependencies. Taking a look at that package I'm on board! Give me a thumbs up and I'll solve it with that.

src/utils.js
0x2b3bfa0: I believe that, if it came escaped, we should include it in the sanitized output, lest corrupting the data.

One thing is removing escape secuences and a very different one is removing printable characters that don't belong to one of such sequences. 😈

0x2b3bfa0: Off–topic, for the laughs: https://en.wikipedia.org/wiki/Constrained_writing

.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     SharedArrayBuffer: 'readonly'
   },
   parserOptions: {
-    ecmaVersion: 2018
+    ecmaVersion: 2020
```
0x2b3bfa0, line 14:
```suggestion
    ecmaVersion: 2018
```

package.json
```diff
@@ -83,6 +83,7 @@
     "pseudoexec": "^0.2.0",
     "semver": "^7.3.5",
     "simple-git": "^2.46.0",
+    "strip-ansi": "^7.0.1",
```
0x2b3bfa0, line 86:
```suggestion
    "strip-ansi": "^6.0.1",
```

src/utils.js
```diff
@@ -6,11 +6,16 @@ const forge = require('node-forge');
 const NodeSSH = require('node-ssh').NodeSSH;
 
 const exec = async (command) => {
+  const { default: stripAnsi } = await import('strip-ansi');
```
0x2b3bfa0, line 9:
```suggestion
const NodeSSH = require('node-ssh').NodeSSH;
const stripAnsi = require('strip-ansi');

const exec = async (command) => {
```

0x2b3bfa0 (commented):
> [_is this how to ESM?_ 🤔](https://github.com/iterative/cml/pull/818/commits/2aee1613cc16e630be8f5d88e51baf4a213d4e1f)

this is how to ESM! :hear_no_evil:

0x2b3bfa0 (dismissed):
Pinned [`strip-ansi`](https://github.com/chalk/strip-ansi) to [v6.0.1](https://github.com/chalk/strip-ansi/releases/tag/v6.0.1), just before https://github.com/chalk/strip-ansi/commit/7cda68dcadde18b19bfa31b6223e9f0e60b3e319 to avoid dealing with ESM imports.

dacbd:
Going to push for the package-lock

dacbd:
nvm :)

0x2b3bfa0 (approved):

PR #856, opened 2021-12-28
Pinned version of dvc

DavidGOrtega:
- Fixes temporarily #855 
 - opens #857


0x2b3bfa0:
Maybe it's more elegant here?

https://github.com/iterative/cml/blob/68e29bf0521079e447a56662852067f13b7d20db/.github/workflows/test-deploy.yml#L144

DavidGOrtega:
We should not spend time on this. is opening indeed another issue to revert this

0x2b3bfa0 (approved):

PR #865, opened 2022-01-10
Fix magic path

0x2b3bfa0:
Would close https://github.com/iterative/setup-cml/issues/56; still doesn't work because of an esoteric path issue involving Windows, Emscripten, `libmagic` and a pinch of salt.

* Upstream fix: [1.3.12](https://github.com/npcz/magic/releases/tag/v1.3.12) https://github.com/npcz/magic/issues/10

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -28,11 +28,6 @@ const exec = async (command) => {
 
 const mimeType = async (opts) => {
   const { path, buffer } = opts;
-  FileMagic.magicFile = PATH.join(
-    __dirname,
-    '..',
-    'node_modules/@npcz/magic/dist/magic.mgc'
-  );
   FileMagic.defaulFlags = MagicFlags.MAGIC_PRESERVE_ATIME;
   const fileMagic = await FileMagic.getInstance();
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,10 +11,11 @@
       "dependencies": {
         "@actions/core": "^1.6.0",
         "@actions/github": "^4.0.0",
-        "@npcz/magic": "^1.3.11",
+        "@npcz/magic": "^1.3.12",
         "@octokit/core": "^3.5.1",
         "@octokit/plugin-throttling": "^3.5.2",
         "@octokit/rest": "18.0.0",
+        "colors": "1.4.0",
         "ec2-spot-notification": "^2.0.3",
         "form-data": "^3.0.1",
         "fs-extra": "^9.1.0",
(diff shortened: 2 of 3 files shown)


0x2b3bfa0:
Currently works everywhere, except when `magic.mgc` is under the drive [`C:`](https://en.wikipedia.org/wiki/Drive_letter_assignment#Common_assignments) 🙃

src/utils.js
```diff
@@ -28,11 +28,7 @@ const exec = async (command) => {
 
 const mimeType = async (opts) => {
   const { path, buffer } = opts;
-  FileMagic.magicFile = PATH.join(
-    __dirname,
-    '..',
-    'node_modules/@npcz/magic/dist/magic.mgc'
-  );
+  FileMagic.magicFile = require.resolve("@npcz/magic/dist/magic.mgc");
```
0x2b3bfa0, line 31:
```suggestion
  FileMagic.magicFile = require.resolve('@npcz/magic/dist/magic.mgc');
```

DavidGOrtega:
> Currently works everywhere, except when `magic.mgc` is under the drive [`C:`](https://en.wikipedia.org/wiki/Drive_letter_assignment#Common_assignments) 🙃

Why? The code is currently OS agnostic right? I mean 

```js
FileMagic.magicFile = PATH.join(
    __dirname,
    '..',
    'node_modules/@npcz/magic/dist/magic.mgc'
  );
```

0x2b3bfa0:
> _The code is currently OS agnostic right? I mean_
> ```javascript
> FileMagic.magicFile = PATH.join(
>     __dirname,
>     '..',
>     'node_modules/@npcz/magic/dist/magic.mgc'
>   );
> ```

If you mean using [`path.join`](https://nodejs.org/api/path.html#pathjoinpaths) just to pass a bunch of forward slashes as one of the arguments... 😈 Not quite agnostic.

0x2b3bfa0:
If any, the agnostic alternative would be this:

```javascript
FileMagic.magicFile = path.join(
  __dirname,
  '..',
  'node_modules',
  '@npcz',
  'magic',
  'dist',
  'magic.mgc'
);
```

I'm testing it also, just in case.

DavidGOrtega:
i think in windows that does not matter. Like python

0x2b3bfa0:
😅 Today I learned that theay are fully equivalent, and you can even mix forward slashes with backslashes in the same path. 🤦🏼‍♂️

0x2b3bfa0:
This works as expected. Somebody is stripping the drive letter from paths.

```batch
cd C:\
vl2png /d/a/setup-cml/setup-cml/assets/vega-lite.json | cml publish
```

DavidGOrtega:
I think that we should close this. In any case a similar code to this one is already merged in prod
@0x2b3bfa0

0x2b3bfa0:
Yes, the problem is still not fixed, but changes on this pull request are no longer relevant

Issue #808, opened 2021-11-09
`cml runner` early shutdown from idle-timeout with active job


dacbd:
Cloud provider: `gcp`
SCM / CICD: `github` / `actions`

I have been playing with short `--idle-timeout` values and I was a bit baffled when I encountered this. One workflow shutdown earlier before the GitHub action completes the final job, and the other completes as expected, but they are using nearly identical workflows.

Successful workflow:
```yml
name: Build and run Inference - DEV
on:
  push:
    branches: [dev]
jobs:
  build:
    environment: dev
    runs-on: ubuntu-latest
    steps:
      - uses: dacbd/gcr-build-push@main
        with:
          tags: latest
          project-id: ${{ secrets.GCP_PROJECT_ID }}
          GCR-key: ${{ secrets.GCP_GCR_KEY }}
          container-name: ***
  deploy-runner:
    runs-on: ubuntu-latest
    environment: dev
    steps:
      - uses: iterative/setup-cml@v1
      - uses: actions/checkout@v2
      - name: Deploy runner on GCP
        env:
          GOOGLE_APPLICATION_CREDENTIALS_DATA: ${{ secrets.GCP_CML_RUNNER_KEY }}
        run: |
          cml-runner \
            --single \
            --idle-timeout=360 \
            --token=${{ secrets.PAT_DCB }} \
            --cloud=gcp \
            --cloud-region=us-west \
            --cloud-type=e2-standard-16
  run-inference:
    needs: [build, deploy-runner]
    runs-on: [self-hosted, cml]
    environment: dev
    steps:
      - uses: actions/checkout@v2
      - name: Setup Container image name
        run: echo "ML_CONTAINER=gcr.io/${{ secrets.GCP_PROJECT_ID }}/***" >> "${GITHUB_ENV}"
      - name: Docker GCR Login
        uses: docker/login-action@v1
        with: 
          registry: gcr.io
          username: _json_key
          password: ${{ secrets.GCP_GCR_KEY }}
      - name: Load container from GCR
        run: docker pull ${{ env.ML_CONTAINER }}:latest
      - name: Run Inference
        run: |
          docker run ****
```
Failed workflow:
```yml
name: Build and run Inference - PROD
on:
  schedule: # Saturday at 1400 UTC
    - cron: '0 14 * * 6'
  push:
    branches: [prod]
jobs:
  build:
    environment: prod
    runs-on: ubuntu-latest
    steps:
      - uses: dacbd/gcr-build-push@v1
        with:
[...]

PR #689, opened 2021-08-10
Bullet proof runner termination

DavidGOrtega:
In GH sometimes the log for a new job or job termination does not fails ending up in two scenarios:
 - a runner that runs forever because it believes that it has a job in the pipeline
 - a runner that terminates prematurely because it does not detect the new job

This attempt 
 - For GH checks the runner `busy` status to determine if the runner is still busy or not despite the registered job. If busy tries another round if idle time. If not but it has jobs in the pipeline restarts the jobs and gives another shot.
 - Removes the chrono in favour of just using the setTimeout with the idle time.
 - Simplifies the job queue logic
 - Listens only for exit runner process
 - DRYes the code a bit
 - Smaller destroy window to allow spot instances to terminate properly
 - closes #690 
 - closes #808

PR #905, opened 2022-03-01
Target Specific CML version on created instance.

dacbd:
The default should be the same as the invoking CML version or unset.
Option `hidden` as this value probably shouldn't be set 98%


dacbd:
An Example of the convenience factor this could provide; testing #689 against #808 in a live setting if I could get it to happen twice 🔍

bin/cml/runner.js
```diff
@@ -553,6 +556,12 @@ exports.builder = (yargs) =>
         description: 'Specifies the subnet to use within AWS',
         alias: 'cloud-aws-subnet-id'
       },
+      cmlVersion: {
+        type: 'string',
+        default: require('../../package.json').version,
```
casperdcl, line 561: :eyes: is it safe to assume this works? I doubt it, esp. for the binary distributions

0x2b3bfa0: Yes, even for the binary distributions. Vercel [`pkg`](https://github.com/vercel/pkg) looks for calls to the `require` function and adapts them appropriately to the virtual file system.

0x2b3bfa0: (Tested and confirmed)

bin/cml/runner.js
```diff
@@ -553,6 +556,12 @@ exports.builder = (yargs) =>
...
+        description: 'CML version to load on tpi instance',
```
0x2b3bfa0, line 562:
```suggestion
        description: 'CML version to load on TPI instance',
```

0x2b3bfa0:
Looks great! @dacbd, can you please include this option in `terraform.test.js` snapshot tests and update them?

dacbd:
Will do.

dacbd:
This test failure looks to be unrelated to any of the changes? https://github.com/iterative/cml/runs/5428037240?check_suite_focus=true#step:7:81

0x2b3bfa0:
Yes, it looks like our tests are broken (?) 🔔 @iterative/cml

casperdcl:
Looks like [GL token length change from 20 to 29](https://github.com/iterative/cml/runs/5440254790?check_suite_focus=true#step:7:76)?

DavidGOrtega:
Yes. Its something that I also saw in #689

DavidGOrtega:
@dacbd https://github.com/iterative/cml/blob/266e6126e66b8e5119290b4c3c083cbf71c47c93/src/drivers/gitlab.test.js#L43

dacbd:
> @dacbd
> 
> https://github.com/iterative/cml/blob/266e6126e66b8e5119290b4c3c083cbf71c47c93/src/drivers/gitlab.test.js#L43

thanks, cherry-picked over.

casperdcl (approved):

DavidGOrtega:
Thanks @dacbd !!! ❤️

0x2b3bfa0:
Ditto! ❤️

PR #955, opened 2022-04-13
Fix pull request auto-merge edge cases

0x2b3bfa0:
This pull request tolerates two additional GraphQL errors that were not being taken into account:

* 'Pull request Protected branch rules not configured for this branch',
* 'Pull request is in clean status'

PR #959, opened 2022-04-13
Fix pull request auto-merge edge cases

0x2b3bfa0:
This pull request tolerates two additional GraphQL errors that were not being taken into account:

* 'Pull request Protected branch rules not configured for this branch',
* 'Pull request is in clean status'

Same as #955, closed by GitHub due to a shallow clone issue

### DIFF
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -437,20 +437,24 @@ class Github {
         }
       );
     } catch (err) {
-      if (
-        !err.message.includes("Can't enable auto-merge for this pull request")
-      )
+      const tolerated = [
+        "Can't enable auto-merge for this pull request",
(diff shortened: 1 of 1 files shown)


casperdcl (approved):
if I squint I can see the crazed logic in \`\` vs '' vs ""

0x2b3bfa0:
> _if I squint I can see the crazed logic in `` vs '' vs ""_

It wasn't me, it was the linter, I swear! 😅

* Single quotes are the default
* Double quotes are used when a string contains a ‘single quote’ or an apostrophe, which is the same in the ASCII table
* Backticks are used when strings contain templates; e.g. <code>`this is a ${template}`</code>

PR #963, opened 2022-04-14
Update `node` engine version constraint

0x2b3bfa0:
See https://github.com/iterative/cml/issues/960#issuecomment-1098562960 for context

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "license": "Apache-2.0",
   "main": "index.js",
   "engines": {
-    "node": ">=16.x"
+    "node": ">=16.0.0"
   },
   "bin": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
   "license": "Apache-2.0",
   "main": "index.js",
   "engines": {
-    "node": "<=16.x"
+    "node": ">=16.x"
```
dacbd, line 33: I think we want this to be more explicit.
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines
```suggestion
    "node": ">=16.0.0"
```

dacbd (changes requested):

dacbd (approved):

PR #968, opened 2022-04-17
Try fallback auto-merge when unknown protection

0x2b3bfa0:
This edge case happens when the access token used doesn't have enough permissions to check whether branch protection is enabled or not. See [6337623700](https://github.com/iterative/discord-corpus/runs/6337623700?check_suite_focus=true) for an example.


src/drivers/github.js
```diff
@@ -447,13 +447,20 @@ class Github {
...
+          winston.warn(
+            `Failed to enable auto-merge: Enable the feature in your repository settings: ${settingsUrl}#merge_types_auto_merge. Trying to merge immediately...`
+          );
+        } else {
+          winston.warn(
+            `Failed to enable auto-merge: Set up branch protection and add "required status checks" for branch '${base}': ${settingsUrl}/branches. Trying to merge immediately...`
+          );
+        }
+      } catch (err) {
+        if (!err.message.includes('Resource not accessible by integration')) throw err;
```
0x2b3bfa0, line 461:
```suggestion
        if (!err.message.includes('Resource not accessible by integration'))
          throw err;
```

DavidGOrtega:
@0x2b3bfa0 is this still going on?

casperdcl (approved):
lgtm but it seems like a GH quirk that may change in future? How would we monitor this?

0x2b3bfa0:
> [...] _it seems like a GH quirk that may change in future? How would we monitor this?_

Updated https://github.com/iterative/cml/pull/968#issue-1206260461

PR #1021, opened 2022-05-25
Runner faster termination

### LINKED ISSUE #990: `runner`  error logs, destructure id of `undefined`
Chasing a different error I have finally uncovered the origin of: `Cannot destructure property 'id' of '(intermediate value)' as it is undefined.`

of more commonly seen as:
```
{"level":"error","message":"\tFailed: Cannot destructure property 'id' of '(intermediate value)' as it is undefined."}
```

When the `cml runner` flag `--single` is used, the self-hosted runner is started with `ephemeral` and automatically de-registers itself when its one job is complete so when cml goes through the shutdown process to also tries and de-register itself the GitHub API returns no results because it's already gone....

DavidGOrtega:
The intention of this PR is:

 - avoid having to wait X secs prior to destroy in every scenario. Only if it needs to cleanup TF resources
 - Unregister the runner prior to try to kill the underlying CI runner to not accept more jobs.
 - Fixes #990 with a proper message

### DIFF
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -21,7 +21,15 @@ const shutdown = async (opts) => {
   RUNNER_SHUTTING_DOWN = true;
 
   const { error, cloud } = opts;
-  const { name, workdir = '', tfResource, noRetry, reason } = opts;
+  const {
+    name,
+    workdir = '',
+    tfResource,
+    noRetry,
+    reason,
+    destroyDelay,
+    destroyDelayed
+  } = opts;
   const tfPath = workdir;
 
   const unregisterRunner = async () => {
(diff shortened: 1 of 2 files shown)


bin/cml/runner.js
```diff
@@ -432,7 +432,6 @@ exports.handler = async (opts) => {
     await run(opts);
   } catch (error) {
     await shutdown({ ...opts, error });
-    throw error;
```
0x2b3bfa0, line 435: Why not throw the error so it's displayed on the command-line output?

DavidGOrtega: Are we ever reaching that? if shutdown succeeds process will terminate

0x2b3bfa0: Good point! 🤦🏼‍♂️ Sounds like more “logic to revisit”, but makes sense.

0x2b3bfa0 (approved):

bin/cml/runner.js
DavidGOrtega: I was there because shutdown was meant to throw errors originally but shutdown is now bulletproof

DavidGOrtega: > Good point! 🤦🏼‍♂️ Sounds like more “logic to revisit”, but makes sense.

Im doing that in small PRs to not have a monolithic one

PR #1026, opened 2022-05-27
Integrate `cml publish` with `cml send-comment`

0x2b3bfa0:
## Usage

```bash
tee report.md <<END
# Report
![cat](cat.jpg)
END

cml send-comment --publish --watch report.md &

while sleep 30; do
  curl --location https://thecatapi.com/api/images/get?format=src > cat.jpg
done
```

> **Note**
> Use `npm install --global github:iterative/cml#comment-publish-report` to install this branch

## Behavior

* `--publish` — uploads and replaces all the local paths on `report.md` (e.g. links & images)
  * e.g. `description` becomes `description`
* `--watch` — watches `report.md` and all the local paths it contains
  * i.e. when any of them changes, updates the comment in the forge

### Experimental

* `--trigger-file` — specify a trigger file with the same behavior as DVC checkpoint file-based API[^1]
  * only effective along with `--watch` 

#### Using a trigger file
```bash
cml send-comment --publish --watch --trigger-file=example report.md &

while true; do
  date > report.md # modify the report
  touch example # trigger an update
  while test -f example; do sleep 1; done # wait for the update to finish
done


```

## Pending

- [x] Figure out the consequences of having no synchronization mechanism[^1]
  * Added a primitive file-based trigger, functionally equivalent to the DVC checkpointing file-based API
- [x] Account for cases where `report.md` doesn't yet exist when starting the watcher
- [x] Use a proper file watcher on all the `paths` instead of polling
- [x] https://github.com/paulmillr/chokidar/issues/1224
  * https://github.com/iterative/cml/pull/1026/commits/4c21ff5436c816665745115c91b54df1678284c7 & https://github.com/iterative/cml/pull/1026/commits/ab28d8de0e5c1e0d7650134d73a5ec3ecb316246

## Reverted
- [x] https://github.com/facebook/jest/issues/10662#issuecomment-923850177
  * https://github.com/iterative/cml/pull/1026/commits/e9465da718865b01db687e7ad7c6daa5c152b197
- [x] https://github.com/iterative/cml/issues/1051
  * Not a blocker, just an [eyesore](https://github.com/iterative/cml/blob/5259fc365619a989e120fbe40e9a49fab1adc0b5/src/cml.js#L158-L159)
  
### Questions

* [x] is there **any** use case for calling `cml publish` directly after we implement this?
  + > maybe (e.g. publish for use outside markdown reports?), though probably fine to keep for backward-compat? — @casperdcl dixit
  + > fine to keep hidden, perhaps, as any other deprecated command? — @0x2b3bfa0 respondebat
  + > [`p3-whatever`](https://xkcd.com/2629) — @casperdcl :upside_down_face:
* [x] How does this work outside CI? Are we asking users to run this before using DVCLive?
  + [yes](https://github.com/iterative/cml/pull/1026#discussion_r884759795), but
  + technically this PR is (currently) standalone without anything to do with https://github.com/iterative/dvclive/issues/91. Moving meta discussion to #1036.

[^1]: [As the DVC team may know, a file-based API needs synchronization to avoid all sorts of pitfalls](https://dvc.org/doc/api-reference/make_checkpoint#description); e.g. like lost events, rate limits, corrupted files... et cetera. This makes me question whether we should follow this approach or not.


0x2b3bfa0:
@casperdcl, `cml publish --update` is not necessary, because storage is content-addressed.

0x2b3bfa0:
An intermediate solution to avoid clutter and preserve desirable properties of storage is:

- Generate a UUIDv4 when the daemon starts
- Use it as a prefix for location-based storage

casperdcl:
no, I mean this:

```bash
$ echo one > file
$ ONE_URL=$(cml publish file)
$ curl -I $ONE_URL | head -n1
HTTP/1.1 200 OK
$ echo two > file
$ TWO_URL=$(cml publish file)
warn: CML detected subsequent publish of same filename in same session. Deleting old file.
curl -I $TWO_URL | head -n1
HTTP/1.1 200 OK
$ curl -I $ONE_URL | head -n1
HTTP/1.1 404 Not Found
```

0x2b3bfa0:
> _CML detected subsequent publish of same filename in same session_

Sounds like https://github.com/iterative/cml/pull/1026#issuecomment-1139604345?

0x2b3bfa0:
Related to https://github.com/iterative/dvclive/issues/91#issuecomment-1121147124

src/cml.js
```diff
@@ -159,9 +165,43 @@ class CML {
...
+      };
+
+      await Promise.all(nodes.map(visitor));
+    };
+
+    if (publish)
+      report = String(await remark().use(publishLocalFiles).process(report));
+
+    if (watch)
+      setInterval(() => {
```
daavoo, line 200: Since the CML watcher has to be called before the script that uses dvclive, could CML expose some var to indicate that this has been used, i.e. `CML_WATCH='markdownfile` ?

casperdcl: 'twould be great to have *all* CLI args default values to be overridable by environ (users may want to keep things secret) :)

0x2b3bfa0: If you mean that running `cml ··· --watch` should set an environment variable in the parent process, no, that's technically impossible unless you resort to [non-portable methods](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).

0x2b3bfa0: _Id est, the following is not possible:_

```console
$ printenv CML_WATCH
$ cml ··· --watch
$ printenv CML_WATCH # prints nothing, you expect the path to report.md
```

The easiest portable method is returning an `export` from the command:

```console
$ printenv CML_WATCH
$ $(cml ··· --watch)
$ printenv CML_WATCH # prints the path to report.md as expected
report.md
```

casperdcl: I mean input vars:

```
run: cml pr
env:
  CML_some_opt: ...
```

0x2b3bfa0: Maybe I misinterpreted @daavoo's “expose some var to indicate that this has been used” part of the comment. 🤔

daavoo: > Maybe I misinterpreted @daavoo's “expose some var to indicate that this has been used” part of the comment. 🤔

You didn't. I was asking for that.

> that's technically impossible unless you resort to [non-portable methods](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).

Could CML support and abstract away those non-portable methods, as it does for other features 👼  ? `cml set-env` ?

---

What I would like to have for DVCLive is an unambiguous signal that this feature has been used and, ideally, the path it's pointing to. Right now, I would need to use some proxy to check that 1. `we are in CI context` (I guess `CI` env var is ok`) and 2. `CML is installed`. I would be missing 3. `the user has actually called --watch` and 4. `the target path`.

It could also work if CML supports input vars, as per Casper's last comment, but this would rely on documenting the way we would like to set up the workflow but users could still end up facing some pitfalls

0x2b3bfa0: For 3 & 4 `cml` could create a file on `/tmp` or something (?) 🪓

0x2b3bfa0:
After https://github.com/iterative/cml/pull/1026#discussion_r884495325 and some limitations of the file event-based interface, I wonder if this is the right approach to integrate CML with other tools. 🤔

src/cml.js
casperdcl: > Right now, I could use some "proxy signals" and check that:
> - we are in CI context (I guess CI env var is ok)
> - CML is available.

I was envisioning something different.

1. DVCLive outputs markdown reports
2. CML watches for said reports

without any prior

0. DVCLive looks for CML

0x2b3bfa0: Also known as:

```
cml whatever --watch report.md
python train.py report.md
```

daavoo: > Why should DVCLive need to know CML exists? 

Because markdown reports are not the default report format, `html` is. 

The idea of an extreme happy path is that the user won't need to change anything in the `train.py`  and DVCLive would decide whether or not to generate the `markdown` report.

It is not a deal-breaker, though. We can just rely on properly documenting the behavior and let users handle the report "type"

casperdcl: > DVCLive would decide whether or not to generate [`md` or `html`] So, same code would generate [`html` (local) or `md` (CI)].

Hmm to me this sounds like a p2-nice-to-have or lower, but maybe

daavoo: Agree. it is minor and any of those would work.

casperdcl: updated the OP/description & resolving here in favour of keeping discussion in #1036

src/cml.js
```diff
@@ -140,13 +155,20 @@ class CML {
   }
 
   async commentCreate(opts = {}) {
+    const { remark } = await import('remark');
+    const { visit } = await import('unist-util-visit');
```
DavidGOrtega, line 159: please do the import statements on top

DavidGOrtega (changes requested):

src/cml.js
0x2b3bfa0: I'm afraid that it's not possible

DavidGOrtega: ummm I do not fully understand

0x2b3bfa0: Sorry for the cryptic reply. I mean that those are ESM (asynchronous) imports and the best we can do is having them at the top of an asynchronous function.

* https://github.com/iterative/cml/issues/1051

0x2b3bfa0: Sorry for the cryptic reply. I mean that those are ESM (asynchronous) imports and the best we can do is having them at the top of an asynchronous function.

* https://github.com/iterative/cml/issues/1051

src/cml.js
```diff
@@ -159,9 +181,63 @@ class CML {
...
+    };
+
+    if (publish)
+      report = String(await remark().use(publishLocalFiles).process(report));
+
+    if (watch) {
+      let lock;
+      watcher.add(triggerFile || markdownFile);
+      watcher.on('all', async (event, path) => {
+        if (lock) return;
```
DavidGOrtega, line 225: Im not sure why do you need the lock. Could you please elaborate?

src/cml.js
```diff
@@ -159,9 +181,63 @@ class CML {
...
+          await this.commentCreate({ ...opts, update: true, watch: false });
+          if (event !== 'unlink' && path === triggerFile)
+            await fs.promises.unlink(triggerFile);
+        } catch (err) {
+          winston.warn(err);
+        }
+        lock = false;
+      });
+      winston.info('watching for file changes...');
+      await new Promise((resolve) => resolve); // wait forever
```
DavidGOrtega, line 238: can not be done with a common `while`?

Also, is the watch mechanism UI blocker? Can I launch `send-comment --watch` and then continue?

DavidGOrtega (changes requested):

src/cml.js
0x2b3bfa0: This pull request is not blocked by #1051

0x2b3bfa0: Babel is necessary because:
1. Jest doesn't fully support ESM yet
2. Vercel `pkg` doesn't support ESM at all, and will also have to depend on Babel; see #1051

src/utils.js
```diff
@@ -29,7 +29,6 @@ const mimeType = async (opts) => {
   const { path, buffer } = opts;
   const magicFile = PATH.join(__dirname, '../assets/magic.mgc');
   if (fs.existsSync(magicFile)) FileMagic.magicFile = magicFile;
-  FileMagic.defaulFlags = MagicFlags.MAGIC_PRESERVE_ATIME;
```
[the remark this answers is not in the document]
0x2b3bfa0: Yes, it's affecting this pull request because of https://github.com/paulmillr/chokidar/issues/1224, as explained on https://github.com/iterative/cml/pull/1026#issue-1250268830

src/cml.js
0x2b3bfa0: > _can not be done with a common `while`?_

Do you mean that some kind of infinite loop could be a better way of waiting forever? 🤔

src/cml.js
DavidGOrtega: thats what I say. I would prefer to have included all that code in #1051 and here just the new feature

package.json
```diff
@@ -45,7 +45,7 @@
   "scripts": {
     "lintfix": "eslint --fix ./ && prettier --write '**/*.{js,json,md,yaml,yml}'",
     "lint": "eslint ./",
-    "test": "jest --forceExit",
+    "test": "jest --forceExit --maxWorkers=3",
```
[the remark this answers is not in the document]
DavidGOrtega: > Why do they have to be async?

Why not in the top and async? Even if we use Babel imports should be done at the top


> Because of [facebook/jest#10662 (comment)](https://github.com/facebook/jest/issues/10662#issuecomment-923850177), as explained on [#1026 (comment)](https://github.com/iterative/cml/pull/1026#issue-1250268830)

Its not CML enforcing node 14.x?

src/cml.js
0x2b3bfa0: > _`while(true)`_

But that is [busy waiting](https://en.wikipedia.org/wiki/Busy_waiting) and would block the event loop! 🙀

src/cml.js
DavidGOrtega: When that scenario may occur?
Indeed if two concurrent calls occurs may corrupt the file writing on top. 
I do not see why do we need this really

src/utils.js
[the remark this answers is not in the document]
DavidGOrtega: > Yes, it's affecting this pull request because of [paulmillr/chokidar#1224](https://github.com/paulmillr/chokidar/issues/1224), as explained on [#1026 (comment)](https://github.com/iterative/cml/pull/1026#issue-1250268830)

Where exactly? I can not find that comment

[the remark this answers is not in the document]
0x2b3bfa0: Captura de pantalla 2022-06-09 a las 16 29 16

src/cml.js
DavidGOrtega: No. Thats the common loop in a game

DavidGOrtega: Events still occurs

DavidGOrtega: You can use in any case as a SetInterval which is nicer

0x2b3bfa0: How can I achieve the same effect with `SetInterval` (?)

DavidGOrtega: In any case... the whole code flow is a bit weird

0x2b3bfa0: That isn't functionally equivalent. The current code just pauses the execution; what you propose doesn't pause the execution and prints a text every second. 🙃

src/cml.js
DavidGOrtega: > Why do they have to be async?

Why not in the top and async? Even if we use Babel imports should be done at the top

src/cml.js
DavidGOrtega: Yeah, I do not think adds something really valuable if you think about it

DavidGOrtega (changes requested):
In my opinion there is a drawback with the command and its that we have to launch it in a foreground process. 
When we did the `tensorboard-dev` we wanted to avoid that and I think that here we could do the same... launch a child process and exit

package.json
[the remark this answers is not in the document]
0x2b3bfa0: And what's the difference?

src/utils.js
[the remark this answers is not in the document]
0x2b3bfa0: It's not the case, unless proven otherwise

src/cml.js
0x2b3bfa0: Did you mean [`wait-foerever`](https://www.npmjs.com/package/wait-forever) by any chance? :sweat_smile:

src/utils.js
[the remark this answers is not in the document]
DavidGOrtega: Errrr OK

src/cml.js
0x2b3bfa0: No need to do that twice; doing it _once_ is enough to cause trouble

package.json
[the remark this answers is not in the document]
DavidGOrtega: Its not that only affected by node 12.x?

src/cml.js
0x2b3bfa0: That's precisely what the `wait-forever` package does.

package.json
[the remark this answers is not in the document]
0x2b3bfa0: And they don't run in Node.js 12

src/cml.js
DavidGOrtega: Yes, one or the other, I do not care while we have a easy to understand name

src/cml.js
DavidGOrtega: again... why async?
Why do you do the load async?

DavidGOrtega: I mean

```js
const uuid = require('uuid');
const winston = require('winston');

const { remark } = import('remark');
const { visit } = import('unist-util-visit');

```

0x2b3bfa0: `import` has to be `await`ed

src/cml.js
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1026/commits/a54bd354ecb09358d8e0e7e2ff97c3d29dfa9410

0x2b3bfa0: Note that the previous code looked so dumb because `new Promise(_ => {})` wasn't allowed by the linter.

0x2b3bfa0:
> _In my opinion there is a drawback with the command and its that we have to launch it in a foreground process.
When we did the tensorboard-dev we wanted to avoid that and I think that here we could do the same... launch a child process and exit_

Relying on the shell to background a process is much more flexible, from the user's perspective, than managing it ourselves.

It requires an extra `&` on the workflow, of course, but also allows users to decide how to run the command-

src/cml.js
```diff
@@ -159,9 +182,63 @@ class CML {
...
+
+    if (watch) {
+      let lock;
+      watcher.add(triggerFile || markdownFile);
+      watcher.on('all', async (event, path) => {
+        if (lock) return;
+        lock = true;
+        try {
+          winston.info(`watcher event: ${event} ${path}`);
+          await this.commentCreate({ ...opts, update: true, watch: false });
```
daavoo, line 230: @0x2b3bfa0 Is there a reason to force `update: true`? Could this respect the `--update` flag?

0x2b3bfa0: When using `--watch` it always updates the last comment instead of creating a new comment on every file change. Otherwise, it would clutter badly your pull request. 🤔

0x2b3bfa0: It could respect the update flag, but don't know if it should.

daavoo: > Otherwise, it would clutter badly your pull request

It could be, but there could also be use cases where people want one comment per-update, basically:

> It could honor the update flag, but don't know if it should.

I think we should honor this and let users decide

src/cml.js
```diff
@@ -159,9 +182,63 @@ class CML {
       ? ''
       : '![CML watermark](https://raw.githubusercontent.com/iterative/cml/master/assets/watermark.svg)';
 
-    const report = `${userReport}\n\n${watermark}`;
+    let userReport = testReport;
+    try {
+      if (!userReport)
+        userReport = await fs.promises.readFile(markdownFile, 'utf-8');
```
dacbd, line 188: Im guessing the pre-commit hook changed this because of the line length?
https://github.com/airbnb/javascript#blocks--braces
```suggestion
      if (!userReport) {
        userReport = await fs.promises.readFile(markdownFile, 'utf-8');
      }
```

src/cml.js
0x2b3bfa0: Then, `--update` would be overloaded, meaning both:

#### Meaning 1
Either
* begin the current session by updating the last CML message, or
* begin the current session with a new message

#### Meaning 2
Either
* use a new message for every update, or
* update the same message for every update

src/cml.js
```diff
@@ -159,9 +182,63 @@ class CML {
...
+    } catch (err) {
+      if (!watch) throw err;
+    }
+
+    let report = `${userReport}\n\n${watermark}`;
     const drv = getDriver(this);
 
+    const publishLocalFiles = () => async (tree) => {
```
dacbd, line 196: Not a fan of this function returns a function syntax /off-topic

src/cml.js
```diff
@@ -159,9 +182,63 @@ class CML {
...
+            if (err.code !== 'ENOENT') throw err;
+          }
+        }
+      };
+
+      await Promise.all(nodes.map(visitor));
+    };
+
+    if (publish)
+      report = String(await remark().use(publishLocalFiles).process(report));
```
dacbd, line 220: Same as above?
```suggestion
    if (publish) {
      report = String(await remark().use(publishLocalFiles).process(report));
    }
```

src/cml.js
```diff
@@ -159,9 +182,63 @@ class CML {
...
+      let lock;
+      watcher.add(triggerFile || markdownFile);
+      watcher.on('all', async (event, path) => {
+        if (lock) return;
+        lock = true;
+        try {
+          winston.info(`watcher event: ${event} ${path}`);
+          await this.commentCreate({ ...opts, update: true, watch: false });
+          if (event !== 'unlink' && path === triggerFile)
+            await fs.promises.unlink(triggerFile);
```
dacbd, line 232: see above
```suggestion
          if (event !== 'unlink' && path === triggerFile) {
            await fs.promises.unlink(triggerFile);
          }
```

dacbd (changes requested):
My only real gripe is `require('wait-forever')`

0x2b3bfa0:
@dacbd, with regard to `wait-forever` feel free to take a look to the review comments on https://github.com/iterative/cml/pull/1026#discussion_r893550349 and propose an alternative solution. It feels that this doesn't deserve a dependency, and absolutely not a helper in a dubiously descriptive `utils` file.

src/cml.js
0x2b3bfa0: Yes, it's because of line length. We have this format all over the place.

0x2b3bfa0: Given that this formatting choice is not exclusive to this pull request, I would address it separately. 😅

0x2b3bfa0: Resolving in favor of https://github.com/iterative/cml/issues/596#issuecomment-1153170448

src/cml.js
0x2b3bfa0: Resolving in favor of https://github.com/iterative/cml/pull/1026#discussion_r895175841

src/cml.js
0x2b3bfa0: Resolving in favor of https://github.com/iterative/cml/pull/1026#discussion_r895175841

src/cml.js
0x2b3bfa0: Same. Do you know any better alternative?

dacbd: not one that I think others would like.

dacbd:
@0x2b3bfa0 if the author had any recent activity and is the package did more that one line sure.
```js
const waitForever = () => new Promise(_ => {});
```

package.json
```diff
@@ -83,22 +84,34 @@
...
     "simple-git": "^3.7.1",
     "strip-ansi": "^6.0.1",
     "strip-url-auth": "^1.0.1",
     "tar": "^6.1.11",
     "tempy": "^0.6.0",
     "timestring": "^6.0.0",
+    "turndown": "^7.1.1",
+    "turndown-plugin-gfm": "^1.0.2",
+    "unist-util-visit": "^4.1.0",
+    "wait-forever": "^1.0.0",
```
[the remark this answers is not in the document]
0x2b3bfa0: https://github.com/iterative/cml/pull/1026/#issuecomment-1153168915

0x2b3bfa0:
> _fork it, give credit, and we can require it under a namespace_

Yes, sure! It wasn't enough to have a single `wait-forever` package on NPM to have two. 😅 I finally replaced it with an[other] `utils.js` wart.

dacbd:
I'm all for using cool supported open-source libraries and supporting people's development of such libraries. But I don't want to needlessly introduce a dependency from an inactive source that is a potential supply chain issue that can effortlessly be avoided by literally c/p'ing one line.....

- no known npm publishes in 3 years
- no known commits in 3 years

(which is totally fine in itself... I have my own substantial gap but I also archived my old packages so people know they are unmaintained and not changing)

but the author's own linked [website - luftywiranda.com](http://luftywiranda.com)  lol click at your own risk.....

dacbd (dismissed):
Code LGTM, functionality I'll defer to others. 😁 🥼

src/cml.js
daavoo: @0x2b3bfa0 What is the current behavior without the "file-watcher"? 

Not sure I follow why it would be ambiguous when adding the file watcher.

0x2b3bfa0: Without specifying `--watcher`
* If `--update` is set, it edits the latest CML comment
* If `--update` is not set, it creates a new comment
In this case, the option has a single meaning

0x2b3bfa0: The question is: should `update` only affect the first run, or also all the subsequent watcher events?

src/cml.js
0x2b3bfa0: @casperdcl, did you mean `--trigger-file` by chance?

src/cml.js
casperdcl: Perhaps we should fix https://github.com/iterative/cml/issues/762 first? Something like:

```
cml report {create [--update],watch} <file>
```

0x2b3bfa0: Would that allow users to “reuse an existing comment and keep updating it”?

0x2b3bfa0: > _`cml report watch` would imply update-like functionality_

That's the current behavior and the whole point of this thread 😅

casperdcl: No, the point of this thread is the API `cml send-comment [--update] --watch <file>` makes no sense, it should be

- `cml report watch <file>`,
- or at least `cml send-comment --watch <file>` should raise
  + `error: --watch needs --update`, or
  + `warning: --watch specified so assuming --update`

0x2b3bfa0: As per https://github.com/iterative/cml/issues/888#issuecomment-1152312400, an optional `--update` could make sense, sort of

0x2b3bfa0: > _`cml send-comment --watch <file>` should raise `warning: --watch specified so assuming --update`_

Suspiciously similar to the current behavior, with two exceptions:
- No warning message is being printed
- The `--update` options determines whether to
   - create a new comment and keep updating it, or
   - just use the last CML comment and keep updating it

casperdcl: I see. Well #888 seems to be an urgent docs issue, while [forcing update-like behaviour](https://github.com/iterative/cml/pull/1026#discussion_r894391433) seems best to me.

Allowing no-update[^1] breaks the use case IMO.

[^1]: purely as a PAT work-around

0x2b3bfa0:
@iterative/cml, shall we merge this?

bin/cml/send-comment.js
```diff
@@ -27,6 +25,25 @@ exports.builder = (yargs) =>
         default: 'HEAD',
         description: 'Commit SHA linked to this comment'
       },
+      publish: {
+        type: 'boolean',
+        description:
+          'Upload local files and images linked from the Markdown report'
+      },
+      watch: {
+        type: 'boolean',
```
casperdcl, line 34: since `--update=true` is forced by `--watch`, it doesn't make sense to accept an `--update` flag, right?

```suggestion
        type: 'boolean',
        conflicts: ['update'],
```

bin/cml/send-comment.js
```diff
@@ -27,6 +25,25 @@ exports.builder = (yargs) =>
...
+        description: 'Watch for changes and automatically update the report'
+      },
+      triggerFile: {
```
casperdcl, line 37:
```suggestion
      watchLockFile: {
```

bin/cml/send-comment.js
```diff
@@ -27,6 +25,25 @@ exports.builder = (yargs) =>
...
+        type: 'string',
+        description: 'File used to trigger the watcher',
```
casperdcl, line 39: Trigger how?

```suggestion
        description: 'If specified, --watch will trigger only when the lockfile is present, and will delete the lockfile',
```

bin/cml/send-comment.js
```diff
@@ -27,6 +25,25 @@ exports.builder = (yargs) =>
...
+        hidden: true
+      },
+      native: {
+        type: 'boolean',
+        description:
+          "Uses driver's native capabilities to upload assets instead of CML's storage. Not available on GitHub."
```
casperdcl, line 45:
```suggestion
          "Uses driver's native capabilities to upload assets instead of CML's storage. Not available on GitHub"
```

src/cml.js
```diff
@@ -159,9 +180,70 @@ class CML {
...
+    let report = `${userReport}\n\n${watermark}`;
     const drv = getDriver(this);
 
+    const publishLocalFiles = async (tree) => {
+      const nodes = [];
+
+      visit(tree, ['definition', 'image', 'link'], (node) => nodes.push(node));
+
+      const visitor = async (node) => {
+        if (node.url && node.alt !== 'CML watermark') {
```
casperdcl, line 201: elsewhere we check only for a string `watermark.svg`, not `node.alt == 'CML watermark'`.

Can we haz consistency? Not fussy about which is better.

https://github.com/iterative/cml/blob/3f2aadcb10a995b57f37c82c7daa9eaae4160376/src/cml.js#L168

casperdcl (changes requested):
only had a super quick glance, just checking - is [this "solution to avoid clutter and preserve desirable properties of storage"](https://github.com/iterative/cml/pull/1026#issuecomment-1139616572) or equivalent implemented?

bin/cml/send-comment.js
0x2b3bfa0: Yes, it does; see https://github.com/iterative/cml/pull/1026#discussion_r894300912

0x2b3bfa0: The `--update` flag controls whether to:
- create a new comment and update it on every watcher event (unset), or
- use an existing comment and update it on every watcher event (set)

bin/cml/send-comment.js
0x2b3bfa0: Lock files don't usually trigger actions 🙃

src/cml.js
0x2b3bfa0: It's not about consistency, it's about intent:

1. `watermark.svg` is used to check if a comment was created by CML
2. `node.alt` is used to check if the current Markdown image is the CML watermark

They are not equivalent; the second is used only when traversing the Markdown AST.

bin/cml/send-comment.js
0x2b3bfa0: Moved to https://github.com/iterative/cml/issues/762#issue-1024172986

bin/cml/send-comment.js
0x2b3bfa0: Exactly as stated in the “using a trigger file” section on https://github.com/iterative/cml/pull/1026#issue-1250268830

bin/cml/send-comment.js
casperdcl: Oh, completely missed this. Really? Seems very complex, overloading `--update` to mean two different things.

- `cml send-comment`: comment
- `cml send-comment --update`: update comment
- `cml send-comment --watch`: comment & periodically update
- `cml send-comment --watch --update`: update comment & periodically update

:nauseated_face:

Can we address #762 first before this PR?

`cml report [--pr] {create [--rm-watermark],update,watch [--update]}` seems 10x more natural and easier to explain/justify.

bin/cml/send-comment.js
casperdcl: Need to do #762 before rewording this again :)

src/cml.js
casperdcl: yes exactly, they're incompatible? The following will cause undefined behaviour, right?

1. `echo 'pwned' >> report.md`
2. `echo 'CML watermark' >> report.md`

casperdcl: also, unrelated but...


```suggestion
        if (node.url && fs.existsSync(node.url) && node.alt !== 'CML watermark') {
```

0x2b3bfa0: > yes exactly, they're incompatible? The following will cause undefined behaviour, right?
> 
> 1. `echo 'pwned' >> report.md`
> 2. `echo 'CML watermark' >> report.md`

Yes, our “watermarks” are utterly broken; we should be using a 128+ bits random string and in a consistent way. Feel free to open a separate issue. 👍🏼

0x2b3bfa0: > _fs.existsSync(node.url)_

Why should we require linked files to exist beforehand? 🤔

casperdcl: well at least check `!node.url.startsWith('http')`?

0x2b3bfa0: Yes, we're adding remote links to the watcher list, but it doesn't quite matter for reasons other than performance.

Nitpick: what if users link to creatively named local files?

```
mkdir https:
touch https://file.com
```

src/cml.js
omesser: having `--update` affect the first iteration only (after that - it updates) sounds reasonable, and might be useful. Why does it break the use case @casperdcl ?

casperdcl: I think we can leave this for #762, not a blocker for me.

src/cml.js
0x2b3bfa0: > _We definitely need to guard against existing URLs_

We already do; no link will be replaced unless a local file with the same path exists and can be successfully uploaded.

bin/cml/send-comment.js
casperdcl: update: If you'd prefer to do #762 immediately after this PR please do feel free to resolve this @0x2b3bfa0

0x2b3bfa0: If you're happy to see https://github.com/iterative/cml/pull/1073 merged immediately after this pull request, fine by me. I would push a new commit to https://github.com/iterative/cml/pull/1073 with unified periods and (perhaps) some better descriptions.

src/cml.js
0x2b3bfa0: Unless proven otherwise, the current implementation already behaves as you propose. Feel free to unresolve if there is any edge case where a remote URL is altered.

casperdcl: ah right `https://file.com` will trigger `FileNotFound` ergo do-nothing, works for me :muk:

casperdcl (approved):

bin/cml/send-comment.js
casperdcl: resolving unresolved

bin/cml/send-comment.js
casperdcl: resolving unresolved

src/cml.js
casperdcl: resolving unresolved

src/cml.js
0x2b3bfa0: Watermark issue moved to https://github.com/iterative/cml/issues/1076

PR #1034, opened 2022-05-30
include [ci skip] in pr commit message and title

francesco086:
I would have preferred to first add tests and then change the code, but as I am inexperienced with JS, I had to give up - perhaps you can do it?


DavidGOrtega:
@francesco086 thanks for the contribution!!! ❤️

Issue #1016, opened 2022-05-24
Meta - `runner` process management and tracking.


dacbd:
Several Issues could help be solved by better management of the GHA process.

Not necessarily solving but the following could be helped by this solution.
- https://github.com/iterative/cml/issues/1014
- https://github.com/iterative/cml/issues/1006
- https://github.com/iterative/cml/pull/802
- OOM reaping / crashes
- Oddities in runner log parsing/detecting events

For cloud runners, launch the GHA client as a systemd unit to better control the process / separate it from the `cml` process. Then hook into logs for monitoring/triggering shutdown events etc.

DavidGOrtega:
Im removing #802 from the list. Its already within the #721 epic

Issue #663, opened 2021-07-20
Discuss about the inclusion of SpotNotifier


DavidGOrtega:
Follow up [here](https://github.com/iterative/cml/pull/653#discussion_r673003503)

dacbd:
by its current presence, https://github.com/iterative/cml/blob/4668433d9cef8f792ce4df1417d1682d72464ce6/bin/cml/runner.js#L310-L317 Im going to assume this discussion was resolved else where.

0x2b3bfa0:
Closing this issue blissfully ignores https://github.com/iterative/cml/pull/653#pullrequestreview-710576259, but nobody will complain. 🤷🏼‍♂️

dacbd:
> Closing this issue blissfully ignores [#653 (review)](https://github.com/iterative/cml/pull/653#pullrequestreview-710576259), but nobody will complain. 🤷🏼‍♂️

We can reopen a clearer issue. What needs to be solved?

0x2b3bfa0:
Action point: handle the spot termination notice on the provider side and translate them into a `SIGTERM`

dacbd:
Shouldn't need to translate it to `SIGTERM`, as I understand it both GCP and AWS use the ACPI power button to signal spot termination within 30 seconds and systemd sends `SIGTERM` to everything

dacbd:
if `SIGTERM` isn't enough then the only other option is to poll the instance metadata service every minute, and that is only an option for AWS, for [GCP](https://cloud.google.com/compute/docs/instances/spot#preemption-process) letting systemd handle the ACPI event is the only method.

dacbd:
perhaps adding a specific script for `ExecStop` could help guarantee some spot-required outcome.

0x2b3bfa0:
> Spot instances can [poll](https://en.wikipedia.org/wiki/Polling_(computer_science)) a special internal URL to know if they're going to be terminated soon:
> * _[Amazon Web Services](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-instance-termination-notices.html#termination-time-metadata) — http://169.254.169.254/latest/meta-data/spot/termination-time_
> * _[Google Cloud](https://cloud.google.com/compute/docs/instances/create-use-preemptible#detecting_if_an_instance_was_preempted) — http://metadata.google.internal/computeMetadata/v1/instance/preempted_
[...]

PR #1068, opened 2022-06-19
Runner detect ACPI shutdown

DavidGOrtega:


omesser:
🤔 
Might there be (in some configuration) a significant time difference between the actual spot notification (what we should be listening for, should be the earliest signal) and the *termination* message on the `/var/run/acpid.socket` socket?
If so - we should probably listen on both (asynchronously) to not lose time (and in some scenarios) and fail to persist state if it's big

DavidGOrtega:
@omesser seems that what AWS is doing is to send the ACPI shutdown signal giving 30 secs to dispose it after that

omesser:
@DavidGOrtega  - but don't we get more (and configurable) time with  the spot notification mechanism? it's the same? and in other providers, any idea?

DavidGOrtega:
> @DavidGOrtega - but don't we get more (and configurable) time with the spot notification mechanism? it's the same? and in other providers, any idea?

Yes, we could use the spotnotifer but seems to not be working?!!

dacbd (changes requested):
https://github.com/iterative/cml/pull/1083

dacbd (approved):
Good to go to master

0x2b3bfa0 (commented):
The [systemd service wrapping `cml runner`](https://github.com/iterative/terraform-provider-iterative/blob/7c800a9d178efe11ade71248915166d6b651453c/iterative/resource_runner.go#L366-L374) suffices to send a TERM signal to the process when the instance transitions to the ACPI G2 state. I believe that this pull request solves an [XY problem](https://xyproblem.info/), unless proven otherwise.

I'm not opposed to merging this pull request if it solves a critical issue, but I would suggest, with a lower priority:
1. Reading https://github.com/iterative/cml/issues/663#issuecomment-1159374581 for more information on alternative approaches, like polling for spot instance termination signals through the internal metadata service.
2. Making sure that `cml runner` is able to handle gracefully a TERM signal and doesn't exit until all the cleanup routines run.

DavidGOrtega:
> The [systemd service wrapping `cml runner`](https://github.com/iterative/terraform-provider-iterative/blob/7c800a9d178efe11ade71248915166d6b651453c/iterative/resource_runner.go#L366-L374) suffices to send a TERM signal to the process when the instance transitions to the ACPI G2 state. I believe that this pull request solves an [XY problem](https://xyproblem.info/), unless proven otherwise.
> 
> I'm not opposed to merging this pull request if it solves a critical issue, but I would suggest, with a lower priority:
> 
> 1. Reading [#663 (comment)](https://github.com/iterative/cml/issues/663#issuecomment-1159374581) for more information on alternative approaches, like polling for spot instance termination signals through the internal metadata service.
> 2. Making sure that `cml runner` is able to handle gracefully a TERM signal and doesn't exit until all the cleanup routines run.

This PR not only fixes the spot termination, indeed fixes the error of job restart setting the ids.
And of course you can test that this PR behaves as expected trying to stop a normal runner in AWS and GPC and see that in many cases it does not unregister while listening to the acpid always do it right

DavidGOrtega:
> should not include any logic to deal with termination notices other than a plain SIGTERM

I need to check why TPI suffices with SIGTERM. The runner with SIGTERM (behaving perfectly) is not always triggered. 
Quick test. Start a runner in GH/GCP and stop it. It wont register

0x2b3bfa0:
Thanks! I'll try to reproduce.

0x2b3bfa0:
Unable to reproduce on AWS nor GCP; tried 3 batches of 30 runners on the former, and 10 runners on the latter.

dacbd:
I can say that I was able to see the bad detection behavior once pre-pr on GCP, but I saw also the correct behavior several times, and also struggled to reproduce it. I see think that: https://github.com/iterative/cml/issues/1016 or operating the runner process as a separate service that `cml.service` monitors, is a better solution.

0x2b3bfa0:
### Related
* https://github.com/iterative/cml/issues/663#issuecomment-1159374581
* https://github.com/iterative/terraform-provider-iterative/issues/90#issuecomment-875712365

0x2b3bfa0:
> _I was able to see the bad detection behavior_

Are you sure that it was caused because the `cml` process didn't get a `SIGTERM` or could it be an issue in Node.js or our code handling it?

DavidGOrtega:
> I can say that I was able to see the bad detection behavior once pre-pr on GCP, but I saw also the correct behavior several times, and also struggled to reproduce it.

This coincides with my experience and I have been always saying this... `SIGTERM` `systemd` is not consistent in my experience (and others)

dacbd:
> > _I was able to see the bad detection behavior_
> 
> Are you sure that it was caused because the `cml` process didn't get a `SIGTERM` or could it be an issue in Node.js or our code handling it?

both were from an organic spot termination.

PR #943, opened 2022-04-10
Add support for Bitbucket native uploads 

0x2b3bfa0:
[Bitbucket Downloads](https://support.atlassian.com/bitbucket-cloud/docs/deploy-build-artifacts-to-bitbucket-downloads) is more similar to [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) than to [GitLab Uploads](https://docs.gitlab.com/ee/administration/uploads.html). Not sure if it's “the right way” of storing report images.

> _[Tokens] must have write or admin access [sic]_

---

_Refactored from an unmerged commit I wrote on 2021-09-25 (#930)_

PR #1073, opened 2022-06-21
Introduce subcommands in a backwards-compatible way

0x2b3bfa0:
This pull request renames all the commands (loosely as proposed) in a backwards-compatible way.

Functionality remains unchanged, and the only difference is that `cml <balderdash>` can also be called by using `cml <noun> <verb>` as suggested.

* [x] depends on #1026 (to avoid rebase hell)
* [ ] Closes https://github.com/iterative/cml/issues/762, at least the first stage
* [ ] opens a `cml.dev` docs issue (including for #1026)
* [ ] opens an issue to add `--publish` & `--watch` options to `cml check create`
* [ ] opens an issue to rename the `--driver` global option for consistency with [forge naming](https://github.com/iterative/cml/pull/1073#discussion_r912863460)
* [ ] opens an issue about [`cml ci` -> `cml <noun> <verb>`](https://github.com/iterative/cml/pull/1073#discussion_r952438142)
* [ ] hide `--rm-watermark`?

### Outstanding[^1] ~un~resolved discussions

* [x] [Homonyms for command verbs](https://github.com/iterative/cml/pull/1073#discussion_r952449793)
* [x] [Consistency of `cml ci` with the `cml <noun> <verb>` structure](https://github.com/iterative/cml/pull/1073#discussion_r952438142)
* [x] [Behavior of `cml comment create` and proposal of an `update` verb](https://github.com/iterative/cml/pull/1073#discussion_r951732127)

[^1]: As per https://github.com/iterative/cml/pull/1073#issuecomment-1224095678, there might be some threads that were marked as resolved but no action was taken to address they suggestions they include. Feel free to take a look.


dacbd:
Would `cml ci` forever live in legacy mode?

0x2b3bfa0:
I was toying with the idea of using `cml repository configure` 😅

casperdcl (commented):
P.S. we don't need `rerun-workflow` backward compat because 'twas never documented/released :P

dacbd (changes requested):
`legacy/runner` -> `cml/runner/start` ?

dacbd:
looking at the code it looks like you can only call it as `cml runner start` and there is no legacy mode `cml runner` is that incorrect?

0x2b3bfa0:
That's incorrect, or at least was incorrect the last time I checked. 🤔 As stated on https://github.com/iterative/cml/pull/1073#issuecomment-1163383184, the only breaking change is that calling `cml runner` **without any options** prints the help message.

dacbd (dismissed):
Working on making my testing process a bit better, but for right now I am pretty satisfied with my slightly hacky testing that this is indeed backwards-compatible[^1]

[^1]: at least for my already existing tests of `pr`, `publish`, and `runner`

casperdcl (changes requested):
w.r.t [this](https://github.com/iterative/cml/issues/762#issuecomment-1160250216), I strongly feel we should consistently use distinct terms to refer to different concepts, lest we confound users:

- `runner`s are `launch`ed (i.e. a superset of `start`ed)
- dashboard/integrations like tensorboard are `connect`ed or `publish`ed to... A background daemon monitors & streams(again, NOT `start`ed. Functionality is extremely similar to `cml report watch --publish`... so let's not call it `start`?)
- `workflow`s are `run` (DEFINITELY NOT `start`ed [unless we want to really confuse users](https://github.com/iterative/cml/issues/762#issuecomment-1163570557))

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
-        cml.js publish <asset>                    Upload an image to build a report
-        cml.js rerun-workflow                     Reruns a workflow given the jobId or
-                                                  workflow Id
-        cml.js runner                             Launch and register a self-hosted
-                                                  runner
-        cml.js send-comment <markdown file>       Comment on a commit
-        cml.js send-github-check <markdown file>  Create a check report
-        cml.js tensorboard-dev                    Get a tensorboard link
+        cml.js check              Manage continuous integration checks
+        cml.js pr <glob path...>  Manage pull requests
```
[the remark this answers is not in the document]
0x2b3bfa0: Not going to happen because of the so called backwards compatibility. 😞

bin/cml.js
```diff
@@ -51,13 +51,29 @@ const handleError = (message, error) => {
...
     description: 'Maximum log level',
     coerce: (value) => configureLogger(value) && value,
     choices: ['error', 'warn', 'info', 'debug'],
     default: 'info'
+  },
+  driver: {
+    type: 'string',
+    choices: ['github', 'gitlab', 'bitbucket'],
+    description:
+      'Forge where the repository is hosted. If not specified, it will be inferred from the environment'
```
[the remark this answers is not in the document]
0x2b3bfa0: I'll take "unspecified" and final periods as global review suggestions.

bin/cml.js
```diff
@@ -51,13 +51,29 @@ const handleError = (message, error) => {
...
+  },
+  repo: {
+    type: 'string',
+    description:
+      'Repository. If not specified, it will be inferred from the environment'
```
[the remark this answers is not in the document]
0x2b3bfa0: The latter. 💯

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
+        cml.js report             Manage reports
+        cml.js repository         Manage repository settings
+        cml.js runner             Manage continuous integration self-hosted runners
+        cml.js tensorboard        Manage tensorboard.dev agents
```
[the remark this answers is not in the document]
0x2b3bfa0: Not sure, although makes sense to me. As per the [getting started](https://www.tensorflow.org/tensorboard/get_started) guide, it's a _tool_.

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
+        cml.js workflow           Manage continuous integration workflows
```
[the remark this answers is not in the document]
0x2b3bfa0: Maybe leave the "runs" part out in case we want to extend functionality later?

bin/cml/attachment.js
```diff
@@ -0,0 +1,8 @@
+exports.command = 'attachment';
```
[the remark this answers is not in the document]
0x2b3bfa0: It's not an asset nor an art[ie]fact; those terms have a well-defined meaning in forges.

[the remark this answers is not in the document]
0x2b3bfa0: It's hidden anyway, but why do you dislike _attachment_ as the noun?

bin/cml/repository.js
```diff
@@ -0,0 +1,8 @@
+exports.command = 'repository';
```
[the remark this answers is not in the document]
0x2b3bfa0: How does `ci` describe the noun that users are interacting with? That command is all about performing Git operations.

0x2b3bfa0:
> _dashboard/integrations like tensorboard are `connect`ed or `publish`ed to... A background daemon monitors & streams(again, NOT `start`ed. Functionality is extremely similar to `cml report watch --publish`... so let's not call it `start`?)_

To my mind, integrations like Tensorboard use daemons/agents/clients/servers/... which have to be *cough* started in order to communicate information to the user. Maybe using the verb `start` everywhere is too reductionist, but having different verbs for every command whose function is equivalent introduces an unneccessary strain in users.

> _**Start** TensorBoard through the command line or within a notebook experience._
> (Excerpt from Tensorboard's [Get **Start**ed](https://www.tensorflow.org/tensorboard/get_started) Guide)

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
-                                                  specified files
-        cml.js publish <asset>                    Upload an image to build a report
-        cml.js rerun-workflow                     Reruns a workflow given the jobId or
-                                                  workflow Id
-        cml.js runner                             Launch and register a self-hosted
-                                                  runner
-        cml.js send-comment <markdown file>       Comment on a commit
-        cml.js send-github-check <markdown file>  Create a check report
-        cml.js tensorboard-dev                    Get a tensorboard link
+        cml.js check              Manage continuous integration checks
```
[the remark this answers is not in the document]
casperdcl: indeed. My intention is more like:

```suggestion
        cml.js check              Manage CI checks (reports-with-statuses)
```

bin/cml.js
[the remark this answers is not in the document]
casperdcl: Yes?

- We don't use `forge` in `cml.dev`
- Everyone knows what a (web)site is

bin/cml/attachment.js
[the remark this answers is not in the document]
casperdcl: > It's not an asset nor an art[ie]fact

orly?

`asset.cml.dev`

casperdcl:
my tuppence: `start`ing daemons is an implementation detail users need not care about.

bin/cml.js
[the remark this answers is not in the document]
0x2b3bfa0: * We don't use the term `forge` in [cml.dev](https://cml.dev) because of an historical vacancy in our team's vocabulary. After joining, I noticed that in all the meetings and conversations GitHub, GitLab and Bitbucket were being erroneously called SCMs; i.e. what Git is, but forges aren't. 🐲 

* [XKCD 1133](https://xkcd.com/1133/); where do we draw the line? Should we use overly generic terms to replace more accurate ones if we suspect than the latter are (allegedly) less known?

bin/cml/attachment.js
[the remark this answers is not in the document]
0x2b3bfa0: Yes, really. 🦉

The `asset.cml.dev` domain was named in an overly generic way[^1] due to historical reasons that are unknown to me.

There is a considerable absence of precise terminology to describe this functionality; e.g. GitLab [uploads](https://docs.gitlab.com/ee/api/projects.html#upload-a-file) (file) or Bitbucket [downloads](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-downloads/#api-repositories-workspace-repo-slug-downloads-post) (artifact) which was added in [2009](https://bitbucket.org/blog/new-feature-downloads) with a functionality similar to ours.

At the end, this functionality is being used to attach files (mostly images) to issues, pull requests and checks. In fact, that's precisely what GitHub displays on the user interface: _"attach files by dragging & dropping, selecting or pasting them"_

Unless we plan to encourage users to use our storage for uploading build artifacts, it would be better to avoid the word artifact. As well, asset is not as concise and user-oriented as attachment.

Anyhow. Just bikeshedding, given that this command won't be directly consumed by users, and it's just an internal tool to *cough* attach plots to reports.

[^1]: https://en.wikipedia.org/wiki/Digital_asset

bin/cml/pr.js
```diff
@@ -1,89 +1,13 @@
...
 exports.command = 'pr <glob path...>';
-exports.description = 'Create a pull request with the specified files';
-
-exports.handler = async (opts) => {
-  const cml = new CML(opts);
-  const link = await cml.prCreate({ ...opts, globs: opts.globpath });
-  if (link) console.log(link);
-};
-
+exports.description = 'Manage pull requests';
```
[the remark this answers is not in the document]
0x2b3bfa0: `./pr/create` describes the `create` verb, while this description is for the `pr` noun.

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
+        cml.js pr <glob path...>  Manage pull requests
+        cml.js report             Manage reports
+        cml.js repository         Manage repository settings
```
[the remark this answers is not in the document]
0x2b3bfa0: Resolving as duplicate of https://github.com/iterative/cml/pull/1073#discussion_r912881350

bin/cml/check/create.test.js
```diff
@@ -36,13 +36,18 @@ describe('CML e2e', () => {
     expect(output).toMatchInlineSnapshot(`
       "cml.js send-github-check <markdown file>
```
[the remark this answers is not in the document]
0x2b3bfa0: Not inherently wrong, just testing an Obsessively Backwards Compatible ™️ alias instead of the new command.

bin/cml.test.js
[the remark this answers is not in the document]
0x2b3bfa0: Resolving in favor of https://github.com/iterative/cml/issues/1090#issue-1297661652

bin/cml/pr.js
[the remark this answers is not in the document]
0x2b3bfa0: Same as
* https://github.com/iterative/cml/pull/1073#discussion_r912877185
* https://github.com/iterative/cml/pull/1073#discussion_r912877386

bin/cml/report/create.js
```diff
@@ -0,0 +1,57 @@
+const kebabcaseKeys = require('kebabcase-keys');
+
+const CML = require('../../../src/cml').default;
+
+exports.command = 'create <markdown file>';
+exports.description = 'Create a report';
```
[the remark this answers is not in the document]
0x2b3bfa0: https://github.com/iterative/cml/pull/1073#discussion_r917435253

bin/cml/report.js
```diff
@@ -0,0 +1,8 @@
+exports.command = 'report';
+exports.description = 'Manage reports';
```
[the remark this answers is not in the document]
0x2b3bfa0: Resolving in favor of https://github.com/iterative/cml/pull/1073#discussion_r917435253

bin/cml/repository.js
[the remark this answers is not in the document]
0x2b3bfa0: Still thinking that the repository concept is better, although as per https://primer.style/cli/foundations/language, `cml repo` is better than `cml repository`

bin/cml.js
```diff
@@ -51,13 +51,29 @@ const handleError = (message, error) => {
   process.exit(1);
 };
 
-const options = {
+exports.options = {
```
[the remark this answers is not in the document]
0x2b3bfa0: Can you provide an example of the difference? 🤔

bin/cml.js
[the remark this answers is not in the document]
casperdcl: - agree that `SCM` is wrong
- agree that `(web)site` is better
- would also like `Git host`
- disagree that `forge` is best. I'd personally find `forge` super confusing and wrong, it's not even jargon - none of the dictionary definitions support a techy use of the word. Note that GitHub self-describes as a "developer platform" while GitLab a "DevOps platform" and Bitbucket "Git based code hosting".

bin/cml.test.js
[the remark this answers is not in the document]
casperdcl: something like this more understandable to users?

```suggestion
        cml.js tensorboard        Manage tensorboard.dev connections
```

bin/cml.js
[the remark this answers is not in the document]
0x2b3bfa0: > _agree that `(web)site` is better_

(Looking up “unilateral agreement” on the dictionary) 😄 

`(web)site` is better than `SCM` in the sense that is not plain wrong, but not better than forge to my mind.

bin/cml/attachment.js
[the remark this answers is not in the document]
casperdcl: as this is an internal command without a public API, the sole consideration should be consistency (i.e. keeping future dev experience straightforward). Ergo we should stick with `asset(.cml.dev)` irrespective of how "overly generic" it may sound.

bin/cml.js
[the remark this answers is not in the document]
casperdcl: current:

```
$ cml pr --help
...
Options:
  --help                 Show help                                     [boolean]
  --version              Show version number                           [boolean]
  --log                  Maximum log level
          [string] [choices: "error", "warn", "info", "debug"] [default: "info"]
  --driver               Forge where the repository is hosted
    [string] [choices: "github", "gitlab", "bitbucket"] [default: infer from the
                                                                    environment]
  --repo                 Repository URL or slug
                                  [string] [default: infer from the environment]
  --token                Personal access token
                                  [string] [default: infer from the environment]
  --md                   Output in markdown format [](url)             [boolean]
...
  --user-name            Git user name         [string] [default: "Olivaw[bot]"]
```

desired:

```
$ cml pr --help
...
Options:
  --md                   Output in markdown format [](url)             [boolean]
...
  --user-name            Git user name         [string] [default: "Olivaw[bot]"]
  --help                 Show help                                     [boolean]
  --version              Show version number                           [boolean]
  --log                  Maximum log level
          [string] [choices: "error", "warn", "info", "debug"] [default: "info"]
  --driver               Forge where the repository is hosted
    [string] [choices: "github", "gitlab", "bitbucket"] [default: infer from the
                                                                    environment]
  --repo                 Repository URL or slug
                                  [string] [default: infer from the environment]
  --token                Personal access token
                                  [string] [default: infer from the environment]
```

casperdcl:
regarding the "overly generic" versus "accurate but not well-known" naming clashes: I'd pick a third and fourth option: "consistent" (stick to our existing docs) and "well-known" (avoid engineering/implementation-specific jargon terms, instead target matching the end-user's intuition)

bin/cml/repository.js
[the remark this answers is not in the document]
0x2b3bfa0: Then, `cml ci fix` it is?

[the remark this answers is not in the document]
0x2b3bfa0: Taking the Git committer identity configuration as a CI fix is also a bit extreme... _from an engineering perspective_.

bin/cml.js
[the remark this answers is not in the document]
omesser: +1 for forge personally.

But we can just say `Forge (Git host provider) ...` to clarify, and unblock this.
Sounds ok?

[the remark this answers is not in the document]
casperdcl: according to poll of @iterative members, it looks like `git provider` or `git host` are 4~5x more popular than `forge`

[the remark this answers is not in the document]
0x2b3bfa0: And the question was _“What generic term for "GitHub/GitLab/Bitbucket" do you think users will understand best?”_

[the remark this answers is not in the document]
0x2b3bfa0: **Top 5 choices of 9**

* `git provider` — 46% (16 voter)
* `git host` — 34% (12 voter)
* `forge` — 9% (3 voter)
* `git forge` — 6% (2 voter)
* `git site` — 3% (1 voter)

bin/cml/pr/create.js
```diff
@@ -0,0 +1,77 @@
...
+exports.command = 'create <glob path...>';
+exports.description = 'Create a pull request with the specified files';
+
+exports.handler = async (opts) => {
+  const cml = new CML(opts);
+  const link = await cml.prCreate({ ...opts, globs: opts.globpath });
+  if (link) console.log(link);
+};
+
+exports.builder = (yargs) => yargs.env('CML_PR').options(exports.options);
```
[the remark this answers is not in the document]
0x2b3bfa0: In theory the `CML_PR` prefix should be enough, but... 🤷🏼

bin/cml/attachment.js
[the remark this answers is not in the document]
DavidGOrtega: our assets does nothing to be with GitLab [uploads](https://docs.gitlab.com/ee/api/projects.html#upload-a-file) nor Bitbucket [downloads](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-downloads/#api-repositories-workspace-repo-slug-downloads-post) at least to my mind.
While the others reflects an object with also a lifecycle (expiring) we do not have that in place.
Assets name comes from the Web dev standard assets where you can store from `.css` to images to `.js`... anything. Normally those files are stored in assets folder or are even served within an asset endpoint normally backed to a CDN to be able to cache it for fast delivery across the world... Every user upload is via pulbish is an asset to be used within the report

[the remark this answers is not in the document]
0x2b3bfa0: > _nor Bitbucket [downloads](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-downloads/#api-repositories-workspace-repo-slug-downloads-post)_

...ish, see #943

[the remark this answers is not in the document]
0x2b3bfa0: Whilst valid, your definition of _asset_ assumes that a responsible developer is in charge of managing them. Doesn't seem to be the case in this scenario, where users upload them and forget forever. 🤔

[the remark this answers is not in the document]
DavidGOrtega: > where users upload them and forget forever. 

Indeed due to the nature of the id (based on file content) and the absence of auth. 

> your definition of asset assumes that a responsible developer is in charge of managing them
I do not see that assumption. Assets is just a folder to upload assets (any kind of file) with size limitation to not exploit us, not having to have a common or known api usage. Use it at your own risk. Users that do not like it wont use it.

DavidGOrtega (changes requested):
Do we pretend to have on file per action for command entry point having the logic in `/src/cml.js` ? 😰 
 -- bin
   -- cml
     -- check
       -- create.js
       -- delete.js

bin/cml.js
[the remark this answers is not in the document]
0x2b3bfa0: Resolving _ad populum_ with 093136ad5925e155889a0e9c2e659546b8566dae

bin/cml/pr/create.js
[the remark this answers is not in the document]
0x2b3bfa0: As per our last meeting, there is no need to specify the subcommand name on environment variables. 🤷🏼‍♂️

bin/cml.js
```diff
@@ -89,13 +89,30 @@ const handleError = (message, error) => {
           description: 'Maximum log level',
           choices: ['error', 'warn', 'info', 'debug'],
           default: 'info'
+        },
```
DavidGOrtega, line 92: This is adding the parameters to commands like `tensorboard`

bin/cml/attachment/publish.js
```diff
@@ -0,0 +1,74 @@
...
+    description: 'MIME type'
+  },
+  file: {
+    type: 'string',
+    alias: 'f',
+    description:
+      'Append the output to the given file or create it if does not exist',
+    hidden: true
+  },
+  repo: {
```
DavidGOrtega, line 69: Is `repo` intentional? in CML now is taken from common parameters

bin/cml/repository/configure.js
```diff
@@ -0,0 +1,33 @@
...
+
+const { GIT_USER_NAME, GIT_USER_EMAIL } = require('../../../src/cml');
+
+exports.command = 'configure';
+exports.description = 'Configure the cloned repository';
+
+exports.handler = async (opts) => {
+  const { cml, telemetryEvent: event } = opts;
+  await cml.ci(opts);
+  await cml.telemetrySend({ event });
```
DavidGOrtega, line 11: this does not exists anymore

bin/cml/tensorboard/start.test.js
```diff
@@ -57,25 +57,29 @@ describe('CML e2e', () => {
...
-                            600 characters.                                   [string]
-            --md            Output as markdown [title || name](url).         [boolean]
-        -t, --title         Markdown title, if not specified, param name will be used.
+            --driver        Git provider where the repository is hosted
+          [string] [choices: \\"github\\", \\"gitlab\\", \\"bitbucket\\"] [default: infer from the
+                                                                          environment]
+            --repo          Repository URL or slug
+                                        [string] [default: infer from the environment]
+            --token         Personal access token
+                                        [string] [default: infer from the environment]
```
DavidGOrtega, line 71: As mentioned above

DavidGOrtega (commented):
There is a lot of repeated boilerplate. 🤢 

```
exports.description = false;
exports.handler = handler;
exports.builder = builder;
```

```
exports.builder = (yargs) =>
  yargs
    .commandDir('something', { exclude: /\.test\.js$/ })
    .recommendCommands()
    .demandCommand()
    .strict();
```

I wonder if it can be reduced somehow...

bin/cml/repository/configure.js
0x2b3bfa0: It **does** exist on the main branch; was it a slip on #1131?

https://github.com/iterative/cml/blob/d6c9a015007ab5ccbcb1fb8fe93d1e32ad36cb81/bin/cml/ci.js#L11

0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/d8833fe55fd0a617a483685f0e3cf2ddd517d22d

DavidGOrtega: Its not needed anymore due to the last refactor and in any case you should not need to invoke it since its going to be invoked by `bin/cml.js`

bin/cml/tensorboard/start.test.js
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/3c05aaa76ba053ebbafbd241e1b8c1b8e1ada3dc

bin/cml/attachment/publish.js
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/3c05aaa76ba053ebbafbd241e1b8c1b8e1ada3dc

bin/cml.js
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/3c05aaa76ba053ebbafbd241e1b8c1b8e1ada3dc

bin/cml/repository.js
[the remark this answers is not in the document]
0x2b3bfa0: @casperdcl & @iterative/cml, how should we name this?

bin/cml/attachment.js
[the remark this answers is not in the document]
0x2b3bfa0: This is worth a whole new issue; unblocking with https://github.com/iterative/cml/pull/1073/commits/bfc9c9dc0ec65258e17860cbec7c98bbf689ad93. 🤷🏼 

See https://github.com/iterative/cml/issues/291#issuecomment-1090652608 (https://github.com/cli/cli/issues/1895) for a detailed discussion on the possible implications of offering this kind of service.

bin/cml.test.js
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/99615060c0c9b8128430573eff1df5f0e291c3e0 🙃

bin/cml/repository/configure.js
DavidGOrtega: this is telling us that we do not have enough integration tests

bin/cml/check/create.test.js
```diff
@@ -36,13 +36,18 @@ describe('CML e2e', () => {
...
 
-      Create a check report
```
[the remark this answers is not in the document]
0x2b3bfa0: No, it shouldn't; tests still use legacy (now hidden) commands; hiding commands implies setting their description to `false`

bin/cml/report/create.js
```diff
@@ -0,0 +1,57 @@
...
+
+exports.handler = async (opts) => {
+  opts.markdownFile = opts.markdownfile;
+  const cml = new CML(opts);
+  console.log(await cml.commentCreate(opts));
+};
+
+exports.builder = (yargs) =>
+  yargs.env('CML_SEND_COMMENT').options(exports.options);
```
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/bde935914c7c6374fab1395bf3b55f9c80d09aba

bin/cml.js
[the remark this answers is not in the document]
0x2b3bfa0: I'm afraid this is not configurable 😅

0x2b3bfa0:
> _Do we pretend to have on file per action for command entry point having the logic in `/src/cml.js`?_ 😰

Yes, unless there is a better alternative, of course.

0x2b3bfa0:
> _There is a lot of repeated boilerplate._

If it were easy to generalize, it would be in a single file. Unfortunately, there are some differences (commands with the same name as the legacy counterpart) that require custom tweaks.

bin/cml.js
```diff
@@ -28,6 +28,26 @@ const setupOpts = (opts) => {
...
+  for (const [oldPrefix, newPrefix] of Object.entries(
+    legacyEnvironmentPrefixes
+  )) {
+    for (const key in process.env) {
+      if (key.startsWith(`${oldPrefix}_`))
+        process.env[key.replace(oldPrefix, newPrefix)] = process.env[key];
+    }
+  }
+
+  console.error(process.env);
```
DavidGOrtega, line 49: intentional?

DavidGOrtega (changes requested):

bin/cml.js
0x2b3bfa0: Yargs! Sorry, definitely not intentional. 🤦🏼‍♂️

0x2b3bfa0:
```suggestion
```

bin/cml.js
```diff
@@ -28,6 +28,25 @@ const setupOpts = (opts) => {
...
+  for (const [oldPrefix, newPrefix] of Object.entries(
+    legacyEnvironmentPrefixes
+  )) {
+    for (const key in process.env) {
+      if (key.startsWith(`${oldPrefix}_`))
+        process.env[key.replace(oldPrefix, newPrefix)] = process.env[key];
+    }
+  }
+
+
```
0x2b3bfa0, line 49:
```suggestion
```

DavidGOrtega (changes requested):
I might be something wrong??
```sh
cml runner start --labels cml --single --repo https://github.com/DavidGOrtega/fashion_mnist --token XXXX 
cml runner

Manage self-hosted (cloud & on-premise) CI runners

Commands:
  cml runner runner  Launch and register a self-hosted runner

Options:
  --help                                    Show help                  [boolean]
  --version                                 Show version number        [boolean]
  --log                                     Maximum log level
          [string] [choices: "error", "warn", "info", "debug"] [default: "info"]
  --driver                                  Git provider where the repository is
                                            hosted
    [string] [choices: "github", "gitlab", "bitbucket"] [default: infer from the
                                                                    environment]
  --repo                                    Repository URL or slug
                                  [string] [default: infer from the environment]
  --token                                   Personal access token
                                  [string] [default: infer from the environment]
  --labels                                  One or more user-defined labels for
                                            this runner (delimited with commas)
                                                       [string] [default: "cml"]
  --idle-timeout                            Time to wait for jobs before
                                            shutting down (e.g. "5min"). Use
                                            "never" to disable
                                                 [string] [default: "5 minutes"]
  --name                                    Name displayed in the repository
                                            once registered
                                                    [string] [default: cml-{ID}]
  --no-retry                                Do not restart workflow terminated
                                            due to instance disposal or GitHub
                                            Actions timeout            [boolean]
  --single                                  Exit after running a single job
                                                                       [boolean]
  --reuse                                   Don't launch a new runner if an
                                            existing one has the same name or
                                            overlapping labels         [boolean]
  --reuse-idle                              Creates a new runner only if the
                                            matching labels don't exist or are
                                            already busy               [boolean]
  --docker-volumes                          Docker volumes, only supported in
                                            GitLab         [array] [default: []]
  --cloud                                   Cloud to deploy the runner
                         [string] [choices: "aws", "azure", "gcp", "kubernetes"]
  --cloud-region                            Region where the instance is
                                            deployed. Choices: [us-east,
                                            us-west, eu-west, eu-north]. Also
                                            accepts native cloud regions
                                                   [string] [default: "us-west"]
  --cloud-type                              Instance type. Choices: [m, l, xl].
                                            Also supports native types like i.e.
                                            t2.micro                    [string]
  --cloud-permission-set                    Specifies the instance profile in
                                            AWS or instance service account in
                                            GCP           [string] [default: ""]
  --cloud-metadata                          Key Value pairs to associate
                                            cml-runner instance on the provider
                                            i.e. tags/labels "key=value"
                                                           [array] [default: []]
  --cloud-gpu                               GPU type. Choices: k80, v100, or
                                            native types e.g. nvidia-tesla-t4
                                                                        [string]
  --cloud-hdd-size                          HDD size in GB              [number]
  --cloud-ssh-private                       Custom private RSA SSH key. If not
                                            provided an automatically generated
                                            throwaway key will be used  [string]
  --cloud-spot                              Request a spot instance    [boolean]
  --cloud-spot-price                        Maximum spot instance bidding price
                                            in USD. Defaults to the current spot
                                            bidding price [number] [default: -1]
  --cloud-startup-script                    Run the provided Base64-encoded
                                            Linux shell script during the
                                            instance initialization     [string]
  --cloud-aws-security-group                Specifies the security group in AWS
                                                          [string] [default: ""]
  --cloud-aws-subnet,                       Specifies the subnet to use within
  --cloud-aws-subnet-id                     AWS           [string] [default: ""]

Unknown argument: start
```

🤔 
```
cml runner runner  Launch
```

DavidGOrtega:
This is working 

```
cml runner runner --labels cml --single --repo https://github.com/DavidGOrtega/fashion_mnist --token XXXX 
```

0x2b3bfa0:
Rebase mistake, fixed with https://github.com/iterative/cml/pull/1073/commits/36c2439fbebce13cba47d7b2d37e71b94cdd9a66 🙈

bin/cml/repository.js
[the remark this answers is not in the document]
dacbd: I think that keeping it as `cml ci` is fine, so is `cml ci fix`. I don't have strong feelings but I am against this being `cml repo/repository` as I feel that has a far different connotation.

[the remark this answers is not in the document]
casperdcl: `cml ci [setup]`?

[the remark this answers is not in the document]
0x2b3bfa0: Note that `cml ci` is already a legacy command that should keep working without arguments. If we overload the old name, running `cml ci` won't show the help menu, but rather “so something” users might not expect.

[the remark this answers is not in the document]
0x2b3bfa0: Omitting this insignificant technical detail, I'm inclined to think that, given that this command modifies the cloned Git repository configuration, `repo` would be a more accurate noun.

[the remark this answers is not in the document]
0x2b3bfa0: Anyway. 🤷🏼‍♂️ Not wanting to make a blocker out of this.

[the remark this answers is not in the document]
DavidGOrtega: Im here with @casperdcl `ci` sounds to me better as its solving a CI case

DavidGOrtega (changes requested):
Silly suggestion.
Would make sense to move legacy.js within the legacy folder?

dacbd (dismissed):
I see no glaring issues. And my set of scheduled tests works with this branch fine.

0x2b3bfa0:
> _Would make sense to move legacy.js within the legacy folder?_

Not much, because
1. `yargs` would try to parse it as a subcommand, and
2. breaks the aesthetics of having matching `.js` files and the pertaining directories[^1]

[^1]: silly reason, originated in the OCD I don't kave ™️

DavidGOrtega:
> 2\. breaks the aesthetics of having matching `.js` files and the pertaining directories[1](#user-content-fn-1-900c210740bded2a4c2921a0f6ec849d)

Its currently breaking my own  OCD aesthetics

bin/cml.js
```diff
@@ -89,13 +107,30 @@ const handleError = (message, error) => {
...
+          type: 'string',
+          defaultDescription: 'infer from the environment',
+          description: 'Personal access token'
         }
       })
       .fail(handleError)
       .middleware(setupOpts)
       .middleware(setupLogger)
       .middleware(setupTelemetry)
-      .commandDir('./cml', { exclude: /\.test\.js$/ })
```
DavidGOrtega, line 98: Do not needed?

bin/cml/pr.js
```diff
@@ -1,79 +1,13 @@
...
-        type: 'string',
-        default: GIT_USER_NAME,
-        description: 'Sets git user name.'
-      }
-    })
-  );
+  yargs
+    .commandDir('./pr', { exclude: /\.test\.js$/ })
+    .recommendCommands()
+    .env('CML_PR')
```
DavidGOrtega, line 10: Why the others do not need it?

DavidGOrtega (changes requested):
I see a bit of noise also in the new boilerplate.

```
exports.builder = (yargs) =>
  yargs
    .commandDir('./report', { exclude: /\.test\.js$/ })
    .recommendCommands()
    .demandCommand()
    .strict();
```

VS 

```
exports.builder = (yargs) =>
  yargs
    .commandDir('./pr', { exclude: /\.test\.js$/ })
    .recommendCommands()
    .env('CML_PR')
    .options(options)
    .check(({ globpath }) => globpath)
    .strict();
```

Originally seemed to work under the same yargs settings

bin/cml.js
0x2b3bfa0: Absolutely needed; otherwise, it won't pick subcommands (in the `cml` directory)

bin/cml/pr.js
0x2b3bfa0: Because `cml pr` already exists as a legacy command, an we can't break compatibility.

bin/cml.js
DavidGOrtega: If needed why is removed? 🤔

bin/cml/pr.js
DavidGOrtega: But the others had the env to right?
Look [this](https://github.com/iterative/cml/pull/1073#pullrequestreview-1079107987)

bin/cml.js
0x2b3bfa0: Oh, you mean just the `exclude` part, not the whole `commandDir` call. 😅 Now the `cml` directory doesn't contain any test files, so there is nothing to exclude.

bin/cml/pr.js
0x2b3bfa0: Others don't need that `.env()` because it was moved to the verbs (sub-sub-commands), but those that share the same name as the legacy commands still need a separate `.env()` methinks.

package.json
```diff
@@ -34,13 +34,13 @@
...
-    "cml-tensorboard-dev": "bin/legacy.js",
-    "cml-runner": "bin/legacy.js",
-    "cml-cloud-runner-entrypoint": "bin/legacy.js",
-    "cml-pr": "bin/legacy.js"
+    "cml-send-github-check": "bin/legacy/link.js",
+    "cml-send-comment": "bin/legacy/link.js",
+    "cml-publish": "bin/legacy/link.js",
+    "cml-tensorboard-dev": "bin/legacy/link.js",
+    "cml-runner": "bin/legacy/link.js",
+    "cml-cloud-runner-entrypoint": "bin/legacy/link.js",
```
DavidGOrtega, line 42: Is this needed? I do not remember this command

DavidGOrtega (approved):
Thanks!!!!
Just a nitpick

package.json
0x2b3bfa0: Yes, this is needed. It's the way we have of retaining compatibility with legacy `cml-command` invocations. When users install the package, `npm` will create symbolic links from e.g. `cml-runner` to `link.js` and the latter will decide what to run based on the link name.

bin/cml/repository.js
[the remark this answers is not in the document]
0x2b3bfa0: From a purist standpoint, the legacy `cml ci` command was almost a shorthand for a fictitious `git config --auto` and, as such, isn't inherently related to continuous integration.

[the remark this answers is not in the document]
0x2b3bfa0: Retaining compatibility for both `cml ci` and `cml ci {verb}` is not possible without sacrificing the help menu. If choosing a name other than `ci` is not possible, we'll have to get very cretive.

package.json
DavidGOrtega: Speaking about
```
"cml-cloud-runner-entrypoint": "bin/legacy/link.js",
```

its that invoked as `cml-cloud-runner`? cml-runner works but not the other

DavidGOrtega: Ah! docker miight be?

0x2b3bfa0: Whatever you did on https://github.com/iterative/cml/pull/487 😅

DavidGOrtega: Ok thats totally deprecated. Lets open a technical debt to remove that

0x2b3bfa0: And break backwards compatibility? 🧛🏼 🧄 😉

0x2b3bfa0: Sounds like something to add under the doubly epic #1090?

DavidGOrtega: 😆 that not breaking anything. Indeed you and casper superseded that with your smart entrypoint later on

0x2b3bfa0:
```suggestion
```

Then, express removal it is.

DavidGOrtega (approved):
🚀 lvgtm!
🙏 thanks for the extra mile!

bin/cml/repository.js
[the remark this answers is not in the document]
0x2b3bfa0: @iterative/cml, this is the only unresolved conversation, let's do something about it.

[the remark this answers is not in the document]
casperdcl: final thought here: because future functionality isn't clear (e.g. `cml aws docker-setup` or `cml config dvc`) let's just leave `cml ci` as-is for now. No need to do `cml ci {verb}` nor `cml repo...`.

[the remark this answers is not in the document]
0x2b3bfa0: Excerpt of my chain of thoughts for the next time we have to take a look at this:

> When this command was introduced we already had some conversations about the description in [cml.dev](http://cml.dev/) because its purpose was rather ethereal.
> 
> * What we say it does: fix the CI system, holistically
> * What we think it does: fix the misdeeds made by different CI/CD systems so repositories behave the same on each of them
> * What it really does: change some configuration options in the cloned repository with git config — shorthand to customize committer name & email, remote authentication, etc.
> 
> Rather cml git config[ure] (noun/verb) but it would be a rather misleading name. :thinking_face:
> 
> I’d stay with `repo config/setup` or similar, but I’d fall into my own trap: repo is also overloaded, meaning both the cloned repository and the remote repository [settings] :exploding_head:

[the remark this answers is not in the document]
0x2b3bfa0: For now, it will remain as `cml ci` without any “modern” counterpart.

bin/cml/report.js
```diff
@@ -0,0 +1,8 @@
+exports.command = 'report';
```
casperdcl, line 1: some options:

1. `cml {report,check} create`: (current) doesn't reflect it's really a comment/check
2. `cml {comment,check} create`: doesn't map well with product description ("CML does **runner**s (`runner` :heavy_check_mark:) and **report**ing (`{comment,check}` :question:)
3. `cml report create [{--comment,--check}]`: a bit long
4. `cml report {comment,check} [--create]`: clunky

dacbd: Not to drag this out more, I agree. `cml report` feels off; cml {comments} a report you make. it doesn't make a report for you.

bin/cml/repository.js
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/f01975ab39b0dca904f636bb8edc9d1aac37c3e1; internal naming & structure follows the command documentation at https://cml.dev/doc/ref/ci for the sake of sanity.

bin/cml/report.js
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/f01975ab39b0dca904f636bb8edc9d1aac37c3e1 by renaming `cml report create` to `cml comment create`

0x2b3bfa0: If we plan to provide a `cml report create` with some higher-level abstraction over comments and checks, it belongs to a separate pull request.

0x2b3bfa0:
🔔 @iterative/cml, I would recommend one last review; not to the code, but to the user experience. E.g. run `cml --help` and navigate the help menus for every subcommand.

bin/cml.js
```diff
@@ -89,13 +107,30 @@ const handleError = (message, error) => {
...
+          description: 'Git provider where the repository is hosted'
+        },
+        repo: {
+          type: 'string',
+          defaultDescription: 'infer from the environment',
+          description: 'Repository URL or slug'
+        },
+        token: {
+          type: 'string',
+          defaultDescription: 'infer from the environment',
```
dacbd, line 124: nitpick:
```suggestion
          defaultDescription: 'infer from the environment ',
```
before: 
image


after:
image

0x2b3bfa0: @dacbd, what's that extra whitespace for? 🤔 To avoid displaying `token[string]` without any whitespace?

0x2b3bfa0: It's probably worthy of a `yargs` issue report rather than a workaround on our side (?)

dacbd: there is no space, `token[string] ...` my ocd would like to see `token [string] ...`

dacbd: obviously this is not a show stopper, very much a nitpick

dacbd: > It's probably worthy of a `yargs` issue report rather than a workaround on our side (?

probably the best option

0x2b3bfa0: Let's see if your OCD is powerful enough to convert the following minimal example into an upstream issue:

```js
require('yargs').default('option', 'description', 'x'.repeat(59)).parse(['--help'])
```

If not, my own OCD will. 😄

bin/cml/comment/create.js
```diff
@@ -0,0 +1,53 @@
...
+    type: 'string',
+    description: 'File used to trigger the watcher',
+    hidden: true
+  },
+  native: {
+    type: 'boolean',
+    description:
+      "Uses driver's native capabilities to upload assets instead of CML's storage; not available on GitHub"
+  },
+  update: {
```
dacbd, line 43: Should there be a `cml comment update` alias `cml comment create --update`?

bin/cml/comment.js
```diff
@@ -0,0 +1,8 @@
+exports.command = 'comment';
```
dacbd, line 1: `cml comment --help` doesn't show options like `cml pr --help` does, both are functionally the same with respect to the `create` subcommand?

bin/cml/pr/create.js
```diff
@@ -0,0 +1,76 @@
+const kebabcaseKeys = require('kebabcase-keys');
+
+const {
+  GIT_REMOTE,
+  GIT_USER_NAME,
+  GIT_USER_EMAIL
+} = require('../../../src/cml');
+
+exports.command = 'create <glob path...>';
```
dacbd, line 9: /note to self `cml pr update` https://github.com/iterative/cml/issues/1126 ?

bin/cml/comment.js
0x2b3bfa0: Thanks to “Obsessive Backwards Compatibility”, some “nouns” like `cml pr` and `cml runner` leak the “verb” options to the parent command. Won't fix without breaking changes. 🙉

bin/cml/pr/create.js
0x2b3bfa0: Probably, but not on this pull request 😅

bin/cml/runner.js
```diff
@@ -1,589 +1,13 @@
...
-    );
-
-  winston.info(`Preparing workdir ${workdir}...`);
-  await fs.mkdir(workdir, { recursive: true });
-  await fs.chmod(workdir, '766');
-
-  if (cloud) await runCloud(opts);
-  else await runLocal(opts);
-};
+const { options, handler } = require('./runner/start');
```
dacbd, line 1: option of `--cloud-aws-subnet-id` printed twice? 
image

bin/cml/pr/create.js
dacbd: > Probably, but not on this pull request

agreed

0x2b3bfa0: Resolving in favor of https://github.com/iterative/cml/issues/1126#issuecomment-1222727199

bin/cml/runner.js
dacbd: `cml runner start --help`

bin/cml/comment.js
dacbd: so what is the correct behavior? to show the "default" subcommand's options or not?

0x2b3bfa0: To **not** show any of the “verb” (child) command in the “noun” (parent) command's help menu. Options should always propagate from parent to child commands (downwards), not upwards.

bin/cml/comment/create.js
0x2b3bfa0: See https://github.com/iterative/cml/issues/762#issuecomment-1163503318 and subsequent comments for more insight on this choice. Won't be part of this pull request, but might make sense.

bin/cml.js
```diff
@@ -89,13 +107,30 @@ const handleError = (message, error) => {
           description: 'Maximum log level',
           choices: ['error', 'warn', 'info', 'debug'],
           default: 'info'
+        },
+        driver: {
```
dacbd, line 111: yargs have a "global option" for these?

for contrast: 
image
vs
image

0x2b3bfa0: Not that I'm aware of, but it would be nice. 😋

bin/cml/comment/create.js
dacbd: maybe we should make a follow-up issue?

0x2b3bfa0: Feel free to 👍🏼

bin/cml.js
0x2b3bfa0: If you mean that `yargs` must have a way of displaying global options separately from local options, I haven't found it yet. 🤔

0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/a9fac4f6be7313fb0e3d20d0e5ca1b2e63844633

bin/cml/comment.js
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/336a3af1d49a8e6e4d1392498fe90589e9691c3b

bin/cml.js
0x2b3bfa0: It did: https://github.com/yargs/yargs/issues/2228

bin/cml/comment/create.js
casperdcl: > See [#762 (comment)](https://github.com/iterative/cml/issues/762#issuecomment-1163503318) and subsequent comments

- deal-breaker for me as per every single "subsequent comment." Definitely not for a follow-up, must be addressed here
  + the primary objective of [#1073](https://github.com/iterative/cml/pull/1073) (<- [#762](https://github.com/iterative/cml/issues/762), [#1026](https://github.com/iterative/cml/pull/1026)) is fixing this particular subcommand's UX. Everything else is comparatively lower-priority technical debt.
- as per https://github.com/iterative/cml/pull/1026#discussion_r901479836, definitely must be one of:
  1. `cml report [--pr,--publish [--native]]`[^native]
     + `create [--rm-watermark]`
     + `update`
     + `watch [{--rm-watermark,--update},--trigger-file]`[^watermark]
  2. `cml report [--pr,--publish [--native],--watch [--trigger-file]]`[^native][^trigger]
     + `create [--rm-watermark]`
     + `update`

Meanwhile, `cml comment create --update` is bad design because it breaks [`CRUD`](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) (`U` is meant to be at the same level as `C`, not a subset). It implies a future with the hilarious `cml comment create --delete` (on par with "[click `start` to shutdown](https://devblogs.microsoft.com/oldnewthing/20030722-00/?p=43083)").

[^watermark]: `--rm-watermark` conflicts with `--update`
[^trigger]: `--trigger-file` requires `--watch`
[^native]: `--native` requires `--publish`

bin/cml/repo.js
```diff
@@ -0,0 +1,8 @@
+exports.command = 'repo';
```
casperdcl, line 1: delete this file? We're sticking with `ci` (not `repo prepare`).

bin/cml/repo/prepare.js
```diff
@@ -0,0 +1,31 @@
+const kebabcaseKeys = require('kebabcase-keys');
```
casperdcl, line 1: delete this file?

bin/cml/repo/prepare.test.js
```diff
@@ -0,0 +1,30 @@
+const { exec } = require('../../../src/utils');
```
casperdcl, line 1: delete this file?

bin/cml/runner/start.js
```diff
@@ -0,0 +1,582 @@
+const { join } = require('path');
```
casperdcl, line 1: still prefer `cml runner launch` to `cml runner start`.

Launching is a superset. It means starting/ignition, configuring/setting up, and taking off. It also removes overloaded/conflicting definition of the word `start` in `cml tensorboard start`. [Overloading (homonyms) are bad for UX](https://tldr.cdcl.ml/arc42-documentation-principles).

bin/legacy/commands/rerun-workflow.js
```diff
@@ -0,0 +1,6 @@
+const { builder, handler } = require('../../cml/workflow/restart');
```
casperdcl, line 1: delete this file. Not needed as per https://github.com/iterative/cml/pull/1073#pullrequestreview-1017307920

bin/legacy/link.test.js
```diff
@@ -1,5 +1,5 @@
-const { bin } = require('../package.json');
-const { exec } = require('../src/utils');
+const { bin } = require('../../package.json');
```
casperdcl, line 1: what's `bin/legacy/link.js` about?

bin/legacy/commands/ci.js
```diff
@@ -0,0 +1,6 @@
+const { builder, handler } = require('../../cml/repo/prepare');
```
casperdcl, line 1: is this meant to be under `/bin/legacy`?

casperdcl (changes requested):
Focusing on meeting the user where they are at. Sensible UX over technical debt/implementation detail. Aiming for [good, future-proof API](https://tldr.cdcl.ml/apis-are-forever) and [avoiding homonyms](https://tldr.cdcl.ml/arc42-documentation-principles).

bin/legacy/link.test.js
0x2b3bfa0: Adapting commands like `cml-send-comment` to the modern non-hyphenated version, like `cml comment`

bin/legacy/commands/ci.js
0x2b3bfa0: Yes, because (1) it's a legacy command and (2) breaks the `cml <noun> <verb>` schema.

bin/cml/comment/create.js
0x2b3bfa0: The `cml <noun> create --update` invocation has a very specific meaning:

1. `cml <noun> create` — create a new `<noun>` or fail if one already exists
2. `cml <noun> update` — update an existing `<noun>` or fail if none was found
3. `cml <noun> create --update` — create a new `<noun>` if none was found, else update an existing one

> **Note**
> To complicate things a bit more, updates are always performed over the CML comment with the latest timestamp, and users have no control over the comment they want to update.

Currently, `cml` always behaves in mode **3**. Official tools like `gh` provide just modes **1** and **2**, relying on the users' judgement to determine whet trusting users whether to create or update and what to update.

0x2b3bfa0: With regard to breaking `CRUD` and actions being at the same level, please note that it's a widespread practice to provide shorthands in command-line tools to merge several verbs into a single invocation.

### Example

#### Verbs
* `command <noun> <verb-1>`
* `command <noun> <verb-2>`

#### Shorthands
* `command <noun> <verb-1> [--verb-2]`
* `command <noun> <verb-2> [--verb-1]`

### References

* [`git checkout -b` — _causes a new branch to be created as if `git-branch` were called and then checked out_](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutem-b-Bltnew-branchgtltstart-pointgt)

0x2b3bfa0: Nitpick: the `--rm-watermark` option should be `--no-watermark` if anything, speaking about user experience.

0x2b3bfa0: > _the primary objective of https://github.com/iterative/cml/pull/1073 (<- https://github.com/iterative/cml/issues/762, https://github.com/iterative/cml/pull/1026) is fixing this particular subcommand's UX_

As per the body of #1073 (this pull request), _“[it] doesn't quite close #762”_

0x2b3bfa0: > _must be addressed here_

Then, I'll start watching beard shaving tutorials.

bin/cml/comment/create.js
```diff
@@ -0,0 +1,53 @@
+const kebabcaseKeys = require('kebabcase-keys');
+
+exports.command = 'create <markdown file>';
+exports.description = 'Create a comment';
```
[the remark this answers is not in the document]
0x2b3bfa0: To my mind, CML messaging about reports has to be backed by a technical `report` feature that simplifies comments and checks. Otherwise, it's going to be rather misleading.

bin/cml/check/create.js
```diff
@@ -0,0 +1,51 @@
...
+
+exports.handler = async (opts) => {
+  const { cml, markdownfile } = opts;
+  const report = await fs.readFile(markdownfile, 'utf-8');
+  await cml.checkCreate({ ...opts, report });
+};
+
+exports.builder = (yargs) => yargs.env('CML_CHECK').options(exports.options);
+
+exports.options = kebabcaseKeys({
```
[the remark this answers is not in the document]
0x2b3bfa0: Please do

bin/cml/comment/create.js
```diff
@@ -0,0 +1,53 @@
+const kebabcaseKeys = require('kebabcase-keys');
+
+exports.command = 'create <markdown file>';
```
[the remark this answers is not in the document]
0x2b3bfa0: That text is also the name of the variable, it would be too long

bin/cml/check/create.js
```diff
@@ -0,0 +1,51 @@
+const fs = require('fs').promises;
+const kebabcaseKeys = require('kebabcase-keys');
+
+exports.command = 'create <markdown file>';
```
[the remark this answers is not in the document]
0x2b3bfa0: That text is also the name of the variable, it would be too long

bin/legacy/commands/rerun-workflow.js
0x2b3bfa0: Absolutely needed as poer https://github.com/iterative/terraform-provider-iterative/blob/823ab139473df13bd0a9fb2adb472941d27b8e27/task/common/machine/machine-script.sh.tpl#L15

0x2b3bfa0:
(To be continued, please check resolved comments just in case some resolution was too heavy-handed; I had to mark them somehow to see clearly in that mess)

bin/cml/comment/create.js
casperdcl: sure, `rm-watermark` could be `no-watermark`, and `update` could be `try-update-fallback-create`. I don't really mind what they're called, open to suggestions. I do mind whether they are `commands` or `--options` though. It's about use cases:

current | proposals (i) or (ii) | use case
---|---|---
`cml comment create --update`| `cml comment update` | one experiment per PR (similar use case to `check`s but uncollapsed report. Similar also to code coverage reports)
`cml comment create` | `cml comment create` | one experiment per commit (multiple experiments per PR)

The `update` use case is far too important to be relegated to an `--option`.

dacbd: I can see the use of the --update flag. ex ci triggered on `pull_request` gets re-run every time the user pushes more commits.

I wouldn't want to the have logic that tries to determine if its the first run to use `create` else `update` additionally I wouldn't want to create multiple comments of essentially the same data with small changes.

I think that `cml comment update` should exist but I agree that something analogous to the `git checkout -b` metaphor should be present.
| example | action |
| --- | --- |
| `cml comment create` | "normal" always creates a comment |
| `cml comment create -u` | looks for watermark -> update else create |
| `cml comment update` | looks for watermark -> updates |
| `cml comment update --id 1234` | tries to update specific comment |

perhaps stealing some HTTP terminology or is that too ugly? `cml comment put`

dacbd:
Possible to change the options ordering so that the "Global Options" appear last?

bin/cml/check/create.js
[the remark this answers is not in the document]
casperdcl: moved to OP checklist

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
+        --log     Maximum log level
+                [string] [choices: \\"error\\", \\"warn\\", \\"info\\", \\"debug\\"] [default: \\"info\\"]
+        --driver  Git provider where the repository is hosted
+          [string] [choices: \\"github\\", \\"gitlab\\", \\"bitbucket\\"] [default: infer from the
+                                                                          environment]
+        --repo    Repository URL or slug[string] [default: infer from the environment]
+        --token   Personal access token [string] [default: infer from the environment]
 
       Options:
         --help     Show help                                                 [boolean]
```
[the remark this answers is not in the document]
tasdomas: Also, is it possible to add a `-h` alias for `--help`?

bin/cml.test.js
```diff
@@ -8,23 +8,26 @@ describe('command-line interface tests', () => {
...
-        --version  Show version number                                       [boolean]
-        --log      Maximum log level
-                [string] [choices: \\"error\\", \\"warn\\", \\"info\\", \\"debug\\"] [default: \\"info\\"]"
+        --version  Show version number                                       [boolean]"
```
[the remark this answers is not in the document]
0x2b3bfa0: While both `--version` and `--help` make more sense as subcommands (syntactically speaking), they are options by convention. It's one of those cases where tradition wins over reason.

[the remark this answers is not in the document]
casperdcl: idk... I increasingly see `help` as a command (e.g. `gh help`, `aws help`). Anyway I don't mind either way (`help` and/or `--help`) but I don't like it appearing in the wrong place. It's either a global option or a subcommand. Not a (local) option, surely?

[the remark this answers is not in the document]
tasdomas: It's something one gets accustomed to.

`cml <subcommand> -h` at least for me seems very natural, but `cml <subcommand> -v` has no valid meaning in most cases.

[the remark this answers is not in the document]
0x2b3bfa0: Yes, `--version` should be a non-global top-level option; it doesn't make sense on subcommands.

bin/cml.test.js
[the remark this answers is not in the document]
0x2b3bfa0: > _shouldn't `--help` be in Global Options?_

Given that `--help` is local to child commands, it probably shouldn't.[^1]

[^1]: Also being lazy because doing such a small change requires an ugly technical workaround.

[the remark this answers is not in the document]
0x2b3bfa0: > _Also, is it possible to add a -h alias for --help?_

Yes, it's possible, although, given that no other option has a single-character shorthand and — as any other unknown option — it displays the help message (e.g. `cml -h` behaves similarly to `cml -xyz` and `cml --help`), I would suggest leaving it as is.

Feel free to resume this thread if you feel it would be useful to add a dedicated alias.

bin/cml/comment/create.test.js
```diff
@@ -16,35 +16,33 @@ describe('Comment integration tests', () => {
     expect(output).toMatchInlineSnapshot(`
       "cml.js send-comment <markdown file>
 
-      Comment on a commit
+      Global Options:
+        --log     Maximum log level
```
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/4077098ecf43ff4b92c1dfa593af3ac6620d8041

bin/cml.test.js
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/b69f3cc2fcd960953317526c6539622a4eac822b

bin/cml.test.js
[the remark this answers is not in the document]
0x2b3bfa0: Nevermind, resolved with https://github.com/iterative/cml/pull/1073/commits/e1f4651c26d8d47b58eed2ae8fe2228973c9bd5f 🤢

0x2b3bfa0:
> _Possible to change the options ordering so that the "Global Options" appear last?_

Still haven't figured out how to do that; assuming it's not possible unless proven otherwise. 🤔

bin/cml/repo.js
0x2b3bfa0: One thing is breaking the command structure for a single command until we find an appropriate noun+verb alternative, and a whole different thing is breaking the project structure because of that. 🤔 If this pull request is “all about user experience” I guess that keeping a consistent project structure won't affect that goal. ⚔️

bin/cml/repo/prepare.js
0x2b3bfa0: https://github.com/iterative/cml/pull/1073#discussion_r957920154

bin/cml/repo/prepare.test.js
0x2b3bfa0: https://github.com/iterative/cml/pull/1073#discussion_r957920154

bin/cml/runner/start.js
0x2b3bfa0: > _~Overloading (homonyms) are bad for UX~_
> [_Glossary: domain/technical terms; avoid synonyms & homonyms to maximise identical understanding_](https://tldr.cdcl.ml/arc42-documentation-principles)

Similarly with synonyms, as per the same page: currently we're in a situation where all the `cml` commands (nouns) have a single subcommand (verb) that performs an action that can be generalized as a `create` action.

While having `create` everywhere may look confusing and repetitive, using unique words to name the same kind of [positive] action isn't any better, to my mind.

bin/cml/workflow/restart.js
```diff
@@ -0,0 +1,18 @@
+const kebabcaseKeys = require('kebabcase-keys');
+
+exports.command = 'restart';
```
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/6429c18a33ac873de58351a88d40ebfcd0b5347f

0x2b3bfa0:
#### Outstanding[^1] unresolved threads

* [Homonyms for command verbs](https://github.com/iterative/cml/pull/1073#discussion_r952449793)
* [Consistency of `cml ci` with the `cml <noun> <verb>` structure](https://github.com/iterative/cml/pull/1073#discussion_r952438142)
* [Behavior of `cml comment create` and proposal of an `update` verb](https://github.com/iterative/cml/pull/1073#discussion_r951732127)

[^1]: As per https://github.com/iterative/cml/pull/1073#issuecomment-1224095678, there might be some threads that were marked as resolved but no action was taken to address they suggestion they include. Feel free to take a look.

bin/cml/repo.js
casperdcl: do you mean `repo prepare` will be (for now) a hidden subcommand (the existence of which is necessary from a technical implementation standpoint)?

bin/cml/comment/create.js
casperdcl: Hmm, taking a step back (ignoring how the choices are specified, i.e. via `subcommand` or via `--option`), this is my reasoning: there are some choices which are nonsensical, as per the table below. Do we agree?

create | update | rm-watermark | watch | behaviour
-|-|-|-|-
1|0|0|0|new
0|1|0|0|mod existing (:warning: personally would prefer "fallback create" as I see zero use case for throwing an error)
1|1|0|0|mod existing, fallback create
1|0|1|0|new, read-only (non-updatable)
0|1|1|0|error
1|1|1|0|mod existing & convert to read-only, fallback create read-only
1|0|0|1|multiple new (:warning: personally would prefer error)
0|1|0|1|mod existing (:warning: personally would prefer "fallback create" as I see zero use case for throwing an error), multiple subsequent mods
1|1|0|1|mod existing, fallback create, multiple subsequent mods
1|0|1|1|error
0|1|1|1|error
1|1|1|1|error

bin/cml/repo.js
0x2b3bfa0: Yes, exactly that.

0x2b3bfa0: Note that the `cml ci` command's logic **could** be placed separately, breaking the `bin/cml/noun/verb.js` and `bin/cml/noun.js` structure without affecting the functionality.

There is not insurmountable technical reason that forces us to preserve the same directory structure for every command, but, to my mind, it simplifies code navigation and reduces cognitive load during development.

casperdcl: > Yes, exactly that.

ok fine by me :)

bin/cml/comment/create.js
0x2b3bfa0: Indeed, some of them are either nonsensical or useless for the workflows we recommend.

To my mind, the following constraints would simplify considerably the workflow:
1. `watch` always implies `update`
2. `update` always implies `create`
3. `--rm-watermark` shouldn't be part of our public API, unless we have a clear use case

0x2b3bfa0: This conversation is probably related to https://github.com/iterative/cml/pull/1073#discussion_r957923322. Currently all of our commands have a single verb, loosely equivalent to the CU[^1] in CRUD.

bin/cml/runner/start.js
casperdcl: > having `create` everywhere may look confusing and repetitive

yes that's a con

> using unique words to name the same kind of [positive] action isn't any better

a) not a con and b) I disagree, it's not at all "the same kind of action"

description | cml runner rerun | cml runner restart
-|-|-
matches term used by GitHub[^1][^2] | :heavy_check_mark: | :x:
matches term used by GitLab[^gl] | :x: | :x:
matches term used by Bitbucket[^bb] | :heavy_check_mark: | :x:
accurately describes behaviour[^desc] | :heavy_check_mark: | :x:

to me, "rerun" is in the same wordcloud as "stop/halt/kill/retry/redo" while "restart" goes with "play/pause/resume"

[^1]: https://cli.github.com/manual/gh_run_rerun
[^2]: https://cli.github.com/manual/gh_workflow_run
[^gl]: https://forum.gitlab.com/t/retry-a-job-multiple-times/58245
[^bb]: https://bitbucket.org/blog/rerun-failed-pipeline-steps
[^desc]: redoing everything from the top of the workflow, i.e. can NOT be mistakenly interpreted to mean "resuming from where it was interrupted"

0x2b3bfa0: @casperdcl, can you please enumerate all the “casperified” names for every verb?

E.g.

| Current verb | New verb |
| ------------ | -------- |
| `cml workflow restart` | `cml workflow rerun` |
| `cml pr created | `cml pr discombobulate` |

bin/cml/comment/create.js
casperdcl: > `watch` always implies `update` [and] `update` always implies `create`

individually, those are fine assumptions. However they can't be combined: there's 2 subtly different use cases for `watch`: a) update existing (no create) and b) create new and keep updating it... meaning `watch` does not always imply `create`.

As I see it, there are 4 use cases, one of which isn't even implemented:

use case | current | proposed (?) | proposed (casper)
-|-|-|-
(1) create comment | `create` | `create` | `create`
(2) update existing, fallback create | `create --update` | `create --update` | `update`
(3) create comment & keep modding it | :warning: not implemented | `create --watch` | `create --watch`
(4) keep modding existing | `create --watch` | `create --watch --update` | `update --watch`
(?) update existing, error if not found | not implemented | `update` | not implemented[^grudge]

Makes sense now? Also [`create --update` seems as illogical to me as `s/workflow rerun/workflow run --rerun/`](https://github.com/iterative/cml/issues/762#issuecomment-1163570557)

> `--rm-watermark` shouldn't be part of our public API

sounds like a follow-up issue; added to OP.

[^grudge]: grudgingly `update --no-create` if users request this functionality

0x2b3bfa0: Isn't 3 implemented? 🤔

bin/cml/runner/start.js
casperdcl: current | proposed (casper)
-|-
workflow restart | workflow rerun
runner start | runner launch

`{pr,comment,check} create` (current) are fine.

0x2b3bfa0: You're missing `tensorboard`

0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/a06918c954aa9e3be5f6bca9e271ff168a409b8b

0x2b3bfa0: Please open a new thread if you want to rename any other verb

0x2b3bfa0: With regard to `restart` versus `rerun`... https://github.com/iterative/cml/pull/1073/commits/6429c18a33ac873de58351a88d40ebfcd0b5347f was pushed 13 days ago. 🤔

bin/cml/comment/create.js
0x2b3bfa0: With regard to (2), isn't it `update --create-if-does-not-exist`? 🤔 Also known as your favorite `update --create` or `create --update` 😅

0x2b3bfa0: Sounds more reasonable to me than `create --delete` or `delete --update` 😸

0x2b3bfa0: Resolved with https://github.com/iterative/cml/pull/1073/commits/aa48409ff0d1e02a3951773cda29aaec54f49362

casperdcl (approved):

casperdcl:
Gratias @0x2b3bfa0!

0x2b3bfa0:
Nihil est, @casperdcl! 🤺 🤝

0x2b3bfa0:
#### Next steps

* @iterative/cml, please make sure that you're happy with the result
* @iterative/scapegoat, please open issues for the remaining items in https://github.com/iterative/cml/pull/1073#issue-1278628641

PR #1092, opened 2022-07-09
optional telemetry

### LINKED ISSUE #1093: CML entrypoint improvements
With #1092 we had to spread the telemetry send command in every CML command. Ideally we should just only have to do it in `bin/cml.js`

### issues

 - We should have only one generic entrypoint in `bin/cml.js` that sets CM, kebabrize parameters, etc...
 - We should never, ever do process.exit outside `bin/cml.js`

DavidGOrtega:
Add optional telemetry to CML to help prioritise features/issues based on usage.

- opens #1093 
- Updated to [JSON spec](https://github.com/iterative/terraform-provider-iterative/pull/622)
- slightly fixes #1093


0x2b3bfa0:
Premature review: consider talking the `deterministic` function from [this Notion page](https://www.notion.so/iterative/User-UUID-generation-0e054a05e717409e82386f17757a2f2a#cee98d7d31b244aaa8785c4b5b9bcdf4).

```javascript
const { promisify } = require('util');
const { scrypt } = require('crypto');
const { v4: uuidv4, v5: uuidv5, parse } = require('uuid');

async function deterministic(data) {
  const namespace = uuidv5('iterative.ai', uuidv5.DNS);
  const name = await promisify(scrypt)(data, parse(namespace), 8, {
    N: 1 << 16,
    r: 8,
    p: 1,
    maxmem: 128 * 1024 ** 2,
  });
  return uuidv5(name.toString('hex'), namespace);
}

function random() {
  return uuidv4();
}

deterministic('example').then((uuid) => {
  console.log(uuid);
  console.log(random());
});
```

DavidGOrtega:
Read desc ☝️

0x2b3bfa0 (dismissed):
A bit too “wet” — please check https://github.com/iterative/cml/pull/1094 for some suggestions on how to avoid duplication.

src/analytics.js
```diff
@@ -0,0 +1,193 @@
...
+  event,
+  endpoint = TPI_ANALYTICS_ENDPOINT,
+  token = TPI_ANALYTICS_TOKEN
+} = {}) => {
+  try {
+    if (ITERATIVE_DO_NOT_TRACK) return;
+    if (!event.id || event.id === ID_DO_NOT_TRACK) return;
+
+    await fetch(endpoint, {
+      signal: timeoutSignal(5 * 1000),
```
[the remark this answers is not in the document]
dacbd: doesn't change that I dislike the bloat.

DavidGOrtega:
> #1094

Nice! Its a good start. I was thinking the same but in another PR because I want to refactor also more things. I think the `cml.js` and subcommands are not very DRY either.

Also it might delay the PR due to testing... but I like the effort... damn... why you just not did wait to the refactor? 😁

0x2b3bfa0 (commented):
Unlike in the Go implementation, there is no code to prevent the program from exiting while telemetry data is still being sent.

Adding `await waitForever()` to the first line of the `telemetry.send` function doesn't prevent the program from exiting. 🤔 Did I overlook something?

DavidGOrtega:
> Unlike in the Go implementation, there is no code to prevent the program from exiting while telemetry data is still being sent.
> 
> Adding `await waitForever()` to the first line of the `telemetry.send` function doesn't prevent the program from exiting. 🤔 Did I overlook something?

I do not follow you @0x2b3bfa0 could you please elaborate? where is the `waitForever`?

src/analytics.js
```diff
@@ -0,0 +1,201 @@
...
+    return {};
+  }
+};
+
+const send = async ({
+  event,
+  endpoint = ITERATIVE_ANALYTICS_ENDPOINT,
+  token = ITERATIVE_ANALYTICS_TOKEN
+} = {}) => {
+  try {
```
0x2b3bfa0, line 176:
```suggestion
  await require('./utils').waitForever();
  try {
```

0x2b3bfa0 (commented):
@DavidGOrtega, I mean that, if you perform the following modification **for testing purposes**, you'll notice that the program exits anyway.

This (allegedly) means that the program won't wait for analytics to finish before exiting.

src/analytics.js
DavidGOrtega: As it is right now? the code will wait until we send the analytics

0x2b3bfa0: How can I test it?

DavidGOrtega:
```js
export ITERATIVE_ANALYTICS_ENDPOINT=http://localhost:3004
export ITERATIVE_ANALYTICS_TOKEN=XXXX
cml publish assets/logo.png
```

DavidGOrtega: my own jitsu

DavidGOrtega: you can setup yours or go against the real one or use an echo server

0x2b3bfa0: But it — obviously — doesn't hang forever. Thus, we can't prove that the `analytics.send` function will return before exiting.

DavidGOrtega: ?

0x2b3bfa0: Short and sweet: the process exits regardless of `analytics.send` being still in progress; if just works because server responses are fast enough.

0x2b3bfa0: Resolving after some out-of-band clarifications: it turns out that I didn't read the latest changes and was commenting while thinking of an outdated version.

DavidGOrtega:
The PR is slow down due to the process.exit and telemetry hook. It needs a bit of thought 😞

bin/cml.js
```diff
@@ -29,59 +61,52 @@ const configureLogger = (level) => {
...
-  RUNNER_SINGLE: 'CML_RUNNER_SINGLE',
-  RUNNER_REUSE: 'CML_RUNNER_REUSE',
-  RUNNER_NO_RETRY: 'CML_RUNNER_NO_RETRY',
-  RUNNER_DRIVER: 'CML_RUNNER_DRIVER',
-  RUNNER_REPO: 'CML_RUNNER_REPO',
-  RUNNER_PATH: 'CML_RUNNER_PATH'
-};
+process.on('uncaughtException', (err) => {
+  handleError('', err, yargs);
+});
```
0x2b3bfa0, line 90: While it works as intended without this modification, I found it slightly mind-bending. Wouldn't it be better to explicitly add `async` and `await`?

```suggestion
process.on('uncaughtException', async (err) => {
  await handleError('', err, yargs);
});
```

bin/cml.js
```diff
@@ -29,59 +61,52 @@ const configureLogger = (level) => {
...
 
-for (const [oldName, newName] of Object.entries(legacyEnvironmentVariables)) {
-  if (process.env[oldName]) process.env[newName] = process.env[oldName];
-}
+process.on('unhandledRejection', (reason) => {
+  handleError('', new Error(reason), yargs);
+});
```
0x2b3bfa0, line 94:
```suggestion
process.on('unhandledRejection', async (reason) => {
  await handleError('', new Error(reason), yargs);
});
```

bin/cml.js
DavidGOrtega: No simply just because it will not do anything at all

0x2b3bfa0: Is it obvious what it does without including the `async` and the `await` for a casual reader?

0x2b3bfa0: I'm not a JavaScript expert, but I struggle to reason an `async` function not being explicitly `await`ed and still acting like if it was. :thin

DavidGOrtega: I can add it but definitely that would be the case for the `casual` reader and not the `js dev` reader. 
Probably like the things I ask you in TPI like explicit returns... Not very golantic way

bin/cml.js
```diff
@@ -29,59 +61,52 @@ const configureLogger = (level) => {
...
+};
+
 const runPlugin = async ({ $0: executable, command }) => {
-  try {
-    if (command === undefined) throw new Error('no command');
-    const path = which.sync(`${basename(executable)}-${command}`);
-    const parameters = process.argv.slice(process.argv.indexOf(command) + 1); // HACK
-    process.exit(await pseudoexec(path, parameters));
-  } catch (error) {
-    yargs.showHelp();
```
dacbd, line 39: Do we know in what scenario this code was being hit, and would we lose without this?
Does the `.fail(handleError)` catch these?

bin/cml.js
```diff
@@ -29,59 +61,52 @@ const configureLogger = (level) => {
...
-    winston.debug(error);
-  }
+  if (command === undefined) throw new Error('no command');
+  const path = which.sync(`${basename(executable)}-${command}`);
+  const parameters = process.argv.slice(process.argv.indexOf(command) + 1); // HACK
+  await pseudoexec(path, parameters);
 };
 
-const handleError = (message, error) => {
+const handleError = async (message, error, yargs) => {
```
dacbd, line 76: what is the difference between message and error/error.message ?

src/cml.js
```diff
@@ -156,6 +157,10 @@ class CML {
     return branch || (await exec(`git branch --show-current`));
   }
 
+  getDriver() {
+    return getDriver(this);
+  }
```
dacbd, line 162: should this be consolidated?

bin/cml/runner.js
```diff
@@ -397,15 +390,22 @@ const run = async (opts) => {
...
+  if (driver === 'github')
     winston.warn(
       'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
     );
-  }
+
+  if (RUNNER_NAME)
+    winston.warn(
+      'ignoring RUNNER_NAME environment variable, use CML_RUNNER_NAME or --name instead'
+    );
```
dacbd, line 408: off-topic, but maybe its time to remove this warning?

dacbd (approved):
lgtm;

bin/cml/runner.js
DavidGOrtega: Might be! There is indeed a collection of vars used in the original containers when we launched the runner via docker.
They might be reviewed

DavidGOrtega: But not here. We should open another issue

bin/cml.js
DavidGOrtega: Yes the handleError will catch it

bin/cml.js
DavidGOrtega: Sorry I do not follow you 🙏

src/cml.js
dacbd: for additional context. https://github.com/iterative/cml/blob/c9d89a44be512c9acd101aa40a91e933baa9dde2/src/cml.js#L79-L88

DavidGOrtega: Consolidated ;)

DavidGOrtega:
Im dismissing @0x2b3bfa0 because I have already his latest golantical suggestions.

DavidGOrtega:
@dacbd I have updated the code can you please review one more time?

report.md
```diff
@@ -0,0 +1 @@
+![logo](assets/logo.png)
```
dacbd, line 1: remove test file?

dacbd (dismissed):
LGTM, this touches enough that I would like to get @0x2b3bfa0 opinion if can prioritize it when he is back.

casperdcl:
(self-requesting review for verifying spec)

src/analytics.js
```diff
@@ -0,0 +1,204 @@
...
+    return {
+      user_id: await userId({ cml }),
+      group_id: await groupId(),
+      action,
+      interface: 'cli',
+      tool_name: 'cml',
+      tool_version: VERSION,
+      tool_source: '',
+      os_name: OS(),
+      os_version: os.release(),
```
casperdcl, line 164: different from Python's `distro.version()`?

casperdcl (commented):
review (mostly comparing to https://github.com/iterative/terraform-provider-iterative/blob/master/iterative/utils/analytics.go)

bin/cml/tensorboard-dev.test.js
```diff
@@ -89,6 +89,10 @@ describe('CML e2e', () => {
         --logdir logs --name '${name}' --description '${desc}'`
     );
 
+    console.log(`node ./bin/cml.js tensorboard-dev --credentials '${CREDENTIALS}' \
```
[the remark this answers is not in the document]
DavidGOrtega: Those credentials are public. Are indeed tin the source code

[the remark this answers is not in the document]
DavidGOrtega: The console.log() has to be removed anyway

src/analytics.js
DavidGOrtega: Definitely different. While some docs states that Im having the version as X.X.X current docs states that we are having the same as uname... For my mac

```
Darwin Kernel Version 19.6.0: Sun Jul  5 00:43:10 PDT 2020; root:xnu-6153.141.1~9/RELEASE_X86_64
```

DavidGOrtega: It the same. In macos prints `19.6.0`

dacbd (dismissed):

0x2b3bfa0:
Still trying to figure out how to avoid injecting telemetry-specific code in every command. What if we wrap the entry point on `bin/cml.js` with some kind of “success” logic?

https://github.com/iterative/cml/blob/6271cb8847bc98b8bf14bc30df5513959896d01f/bin/cml.js#L96-L115

E.g. `yargs...fail(handleError).parse()` can be replaced with `handleSuccess(yargs...fail(handleError).parse())` and, provided that `handleError` contains a `process.exit` they'll be mutually exclusive.

src/analytics.js
casperdcl: clearly not the same:


```bash
$ node
> os.release()
'5.15.0-43-generic'
$ python
>>> import distro
>>> distro.version()
'22.04'
```

DavidGOrtega: @casperdcl 

image

DavidGOrtega: It might be some discrepancies but it should be the same

casperdcl: Well I had differences on a vanilla Ubuntu machine (which is what most CML jobs will be run on) so we need to fix this. Can you test on an ubuntu machine?

0x2b3bfa0 (approved):
Can probably be merged without applying #1124 first. Worth [walking the plank](https://en.wikipedia.org/wiki/Walking_the_plank), though. Yargs!

src/analytics.js
DavidGOrtega: Right now with [a7944f8](https://github.com/iterative/cml/pull/1092/commits/a7944f81505683753c08226ce72bfab049d57792)  on Macos 

```sh
os_name: 'macos',
os_version: '19.6.0',
```

on Ubuntu 
```sh
 os_name: 'linux',
 os_version: '20.04',
```

0x2b3bfa0: For `os_name: 'macos'` the `os_version` is the Darwin Kernel version, not the operating system version.

DavidGOrtega: @0x2b3bfa0 where ? here or in the python version?

0x2b3bfa0: In the comment you posted: https://github.com/iterative/cml/pull/1092#discussion_r942312221

casperdcl: consistency fine by me too :)

casperdcl (commented):
lgtm

DavidGOrtega:
> lgtm

But not approved 🙈

0x2b3bfa0 (approved):
Looks mergeable enough to me

PR #1098, opened 2022-07-21
Detect and warn for `timeout-minutes` change

dacbd:
+/- Provides a paradigm for other checks/warnings in the future
 - Debatably overly complicated detection of relevant timeout-minutes
 
 Sample of warning: 
image

Example of detection:
image

Same file, an example of non-detection:
image

### DIFF
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -401,7 +401,7 @@ const run = async (opts) => {
     }
   }
   console.log(cml.driver);
-  if (driver === 'github') {
+  if (cml.driver === 'github') {
     winston.warn(
       'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
     );
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -400,7 +400,7 @@ const run = async (opts) => {
       process.exit(0);
     }
   }
-
+  console.log(cml.driver);
   if (driver === 'github') {
     winston.warn(
       'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -402,9 +402,6 @@ const run = async (opts) => {
   }
 
   if (cml.driver === 'github') {
-    winston.warn(
-      'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
-    );
     await cml.workflowCheck();
   }
 
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -405,6 +405,7 @@ const run = async (opts) => {
     winston.warn(
       'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
     );
+    await cml.workflowCheck();
   }
 
   winston.info(`Preparing workdir ${workdir}...`);
--- a/bin/cml/runner.js
+++ b/bin/cml/runner.js
@@ -400,7 +400,7 @@ const run = async (opts) => {
       process.exit(0);
     }
   }
-  console.log(cml.driver);
+
   if (cml.driver === 'github') {
     winston.warn(
       'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -3,6 +3,7 @@ const { spawn } = require('child_process');
 const { resolve } = require('path');
 const fs = require('fs').promises;
 const fetch = require('node-fetch');
+const yaml = require('js-yaml');
 
 const github = require('@actions/github');
 const { Octokit } = require('@octokit/rest');
@@ -729,7 +730,6 @@ class Github {
     );
     const WARNINGS = [];
     const _DIR = '.github/workflows/';
-    const yaml = require('js-yaml');
     const files = await fs.readdir(_DIR);
     for (const file of files) {
       const rawStr = await fs.readFile(`${_DIR}${file}`, 'utf8');
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -724,6 +724,9 @@ class Github {
   }
 
   async checkWorkflowForTimeoutMinutes() {
+    winston.warn(
+      'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
+    );
     const WARNINGS = [];
     const _DIR = '.github/workflows/';
     const yaml = require('js-yaml');
--- a/src/drivers/github.js
+++ b/src/drivers/github.js
@@ -735,35 +735,39 @@ class Github {
       const rawStr = await fs.readFile(`${_DIR}${file}`, 'utf8');
       const doc = yaml.load(rawStr);
       Object.keys(doc.jobs).forEach((job) => {
-        // does job contain timeout?
         const timeoutVal = doc.jobs[job]['timeout-minutes'];
-        if (timeoutVal === undefined) return;
-        if (parseInt(timeoutVal) === 50400) return;
-        // does job have label runs-on "self-hosted" or "cml"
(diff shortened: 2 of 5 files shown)


src/drivers/github.js
```diff
@@ -702,6 +702,77 @@ class Github {
...
+          });
+        }
+      });
+    }
+    for (const warning of WARNINGS) {
+      console.log(
+        `::warning file=${warning.file},line=${warning.line},title=Possible Unexpected Behavior using \`cml runner\`::GitHub Actions has updated timeout-minutes if your job is unlikely to longer than 6hrs then you should remove timeout-minutes, to have cml auto-restart for long training jobs change this value to: 50400 (35d)`
+      );
+    }
+    process.exit();
```
DavidGOrtega, line 773: we do not want to do process exit within the sdk/lib. It has to be at most within the CLI so probably should throw an error that will be handled by bin/cml.js

DavidGOrtega (changes requested):

src/drivers/github.js
dacbd: that is only present for spam testing 👍

src/drivers/github.js
```diff
@@ -713,6 +722,72 @@ class Github {
...
+    const _DIR = '.github/workflows/';
+    const yaml = require('js-yaml');
+    const files = await fs.readdir(_DIR);
+    for (const file of files) {
+      const rawStr = await fs.readFile(`${_DIR}${file}`, 'utf8');
+      const doc = yaml.load(rawStr);
+      Object.keys(doc.jobs).forEach((job) => {
+        // does job contain timeout?
+        const timeoutVal = doc.jobs[job]['timeout-minutes'];
+        if (timeoutVal === undefined) return;
```
casperdcl, line 737: well undefined `timeout` is also a problem right?

- timeout 4320: raise
- timeout undefined: raise
- timeout 50400: ok

src/drivers/github.js
```diff
@@ -713,6 +722,72 @@ class Github {
...
+        // does job have label runs-on "self-hosted" or "cml"
+        const runsOn = doc.jobs[job]['runs-on'];
+        switch (typeof runsOn) {
```
casperdcl, line 740: not sure I follow the intention of this switch statement :thinking:

src/drivers/github.js
dacbd: I think there are plenty of use case where cml runner instances last less that 6hrs and `timeout-minutes` is never required. Hence my effort to not bother those users with warning messages :)

src/drivers/github.js
dacbd: it can a string or a list of strings

bin/cml/runner.js
```diff
@@ -401,10 +401,11 @@ const run = async (opts) => {
     }
   }
 
-  if (driver === 'github') {
+  if (cml.driver === 'github') {
     winston.warn(
       'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
     );
```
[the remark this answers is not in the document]
dacbd: Since it's a regular log line I left it. Agreed moving it does make sense though +1

src/drivers/github.js
dacbd: ✔️ `timeout 50400: noop, all ok`

src/drivers/github.js
```diff
@@ -702,6 +702,15 @@ class Github {
     return GITHUB_SHA;
   }
 
+  async workflowCheck() {
+    try {
+      // TODO post telemetery merge - only run if in ci?
+      await this.checkWorkflowForTimeoutMinutes();
+    } catch (err) {
+      // noop
+    }
```
DavidGOrtega, line 711: Why do we need the catch? Can it break by any reason?

src/drivers/github.js
```diff
@@ -713,6 +722,76 @@ class Github {
...
     return 'GitHub Action';
   }
+
+  async checkWorkflowForTimeoutMinutes() {
+    winston.warn(
+      'Github Actions timeout has been updated from 72h to 35 days. Update your workflow accordingly to be able to restart it automatically.'
+    );
+    const WARNINGS = [];
+    const _DIR = '.github/workflows/';
+    const yaml = require('js-yaml');
```
DavidGOrtega, line 732: Should be on top of the file with all the requires

src/drivers/github.js
```diff
@@ -713,6 +722,76 @@ class Github {
...
+        const timeoutVal = doc.jobs[job]['timeout-minutes'];
+        if (timeoutVal === undefined) return;
+        if (parseInt(timeoutVal) === 50400) return;
+        // does job have label runs-on "self-hosted" or "cml"
+        const runsOn = doc.jobs[job]['runs-on'];
+        switch (typeof runsOn) {
+          case 'string':
+            if (
+              runsOn !== 'self-hosted' &&
+              !runsOn.toLocaleLowerCase().includes('cml')
```
DavidGOrtega, line 748: Users are not enforced to set `cml`

src/drivers/github.js
dacbd: it could have weird fs or parsing errors, since this is a "nice-ity" if it has any errors it should not stop any normal operation by raising an error.

src/drivers/github.js
dacbd: Correct, this is an attempt to reduce false positives; I'm airing on the side of being less intrusive on the warnings.

src/drivers/github.js
DavidGOrtega: But following the code I can see it pretty nice... reads the workflow files if any and then parse the yaml, looks nice to me. I can see your fear still so... the only thing is that might not belong to the cml api then because it does not returns or sets anything just only warns

src/drivers/github.js
DavidGOrtega: I think that warning if they set that specific amount of time is a pretty sure asumption

dacbd: if by the specific amount you mean: `timeout-minutes: 4320` I can agree with that argument, but I was explicitly trying to avoid warning simply on the presence of `timeout-minutes` since there are plenty of valid reasons for that to be used.

DavidGOrtega: ok

src/drivers/github.js
dacbd: by "cml api" you mean where this code is living in the project? I decided to do this way so that we could always call `checkworflow` and there is a mechanism to add any other future checks that are forge specific in case we introduce other similar changes which require forge-specific updates. github actions vs gitlab-ci vs bb-pipelines

src/drivers/github.js
dacbd: ✔️ `timeout 4320: raise`
- I disagree on `timeout undefined: raise`

✔️ / ➖ `timeout != 50400: maybe warn?` last timeout condition checked is 4320 / so if:
  1. timeout is set
  2. timeout is not 50400
  3. timeout is not 4320
  4. check for cml usage, then raise

DavidGOrtega (approved):

DavidGOrtega:
stale?

dacbd:
At this point I think its best for this to wait for the command renaming pr

0x2b3bfa0 (commented):
I wrote some review comments but they remain `pending` because I forgot to click the green button :see_no_evil:

bin/cml/runner.js
```diff
@@ -401,10 +401,11 @@ const run = async (opts) => {
     }
   }
 
-  if (driver === 'github') {
+  if (cml.driver === 'github') {
```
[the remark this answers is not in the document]
dacbd: it actually wasn't/isn't and was indeed a bug fix. the previous way would never print unless you explicitly set the driver (ie not infer driver)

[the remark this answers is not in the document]
0x2b3bfa0: 😅 🙏🏼

0x2b3bfa0:
> [_At this point I think its best for this to wait for the command renaming pr_](https://github.com/iterative/cml/pull/1098#issuecomment-1219739433)

* #1073

0x2b3bfa0:
@dacbd, it looks like you'll have to rebase things a bit after #1073

PR #1114, opened 2022-07-31
Fix link cache bypass

### LINKED ISSUE #1110: send-comment `--publish` can not be rendered properly in a browser
Im playing with an experiment case in Bitbucket

```yaml
pipelines:
  default:
    - step:
        image: iterativeai/cml:0-dvc2-base1
        script:
          - |
            cml runner \
                --cloud=aws \
                --cloud-region=us-west \
                --cloud-type=m5.2xlarge \
                --cloud-spot \
                --reuse \
                --labels=cml
    - step:
        runs-on: [self.hosted, cml]
        image: iterativeai/cml:0-dvc2-base1
        # GPU not yet supported, see https://github.com/iterative/cml/issues/1015
        script:
          - |
            npm install -g https://github.com/iterative/cml

            pip install -r requirements.txt
            python get_data.py
            python train.py
            
            # Create pull request
            cml pr model/random_forest.joblib

            # Create CML report
            cat model/metrics.txt > report.md
            echo "\n![confusion_matrix](model/confusion_matrix.png)" >> report.md
            cml send-comment --pr --update --publish report.md
```

The report is not rendering the image. The reason seems to be an added slash to escape the ampersand.
I think that we should URLencode the paramters instead of escaping if its the case.

```
https://asset.cml.dev/3dc922174c992fa28f7cf78645114aa48c61159c?cml=png\&cache-bypass=426fb335-f530-43b2-8a51-8d9afb586a4d
``` 

confusion

In GH seems to work but not in the browser

image

DavidGOrtega:
- closes #1110 
 - Refactor: Removes unused uuid session
 - Refactor: Moves the `cache-bypass` key to publish and utils
 - Refactor: `fs.promises` to `fs`
 - Refactor: `if (updated && nested)` to nested if

The escape backslash is happening within the remark `process` method. While not super elegant the proposed regex replacement I could not find another way of solving it without having to look into remark internals.

WARNING: This solution is adding always the cache bypass and that might be not desirable

### DIFF
--- a/src/cml.js
+++ b/src/cml.js
@@ -4,9 +4,8 @@ const stripAuth = require('strip-url-auth');
 const globby = require('globby');
 const git = require('simple-git')('./');
 const path = require('path');
-const fs = require('fs');
+const fs = require('fs').promises;
 const chokidar = require('chokidar');
-const uuid = require('uuid');
 const winston = require('winston');
 const remark = require('remark');
 const visit = require('unist-util-visit');
@@ -14,7 +13,13 @@ const visit = require('unist-util-visit');
 const Gitlab = require('./drivers/gitlab');
 const Github = require('./drivers/github');
 const BitbucketCloud = require('./drivers/bitbucket_cloud');
-const { upload, exec, watermarkUri, waitForever } = require('./utils');
+const {
+  upload,
+  exec,
+  watermarkUri,
+  preventcacheUri,
+  waitForever
+} = require('./utils');
 
 const { GITHUB_REPOSITORY, CI_PROJECT_URL, BITBUCKET_REPO_UUID } = process.env;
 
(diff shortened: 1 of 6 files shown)


src/cml.js
```diff
@@ -220,11 +219,13 @@ class CML {
...
     if (publish) {
-      report = String(
+      report = (
         await remark()
           .use(() => publishLocalFiles)
           .process(report)
-      );
+      )
+        .toString()
+        .replace(/\\&(.+)=/g, '&$1=');
```
dacbd, line 228: I am unfamiliar with syntax for referencing of the capture groups so assuming this is correct that

`\&something=` would be `$something=`

then +1

dacbd (approved):

PR #1121, opened 2022-08-05
improper terraform ref causes `Extraneous JSON` error

dacbd:
The correct terraform field is `aws_subnet_id` but it was being passed as `aws_subnet`

### DIFF
--- a/src/terraform.js
+++ b/src/terraform.js
@@ -77,43 +77,47 @@ const iterativeProviderTpl = ({ tpiVersion }) => ({
   }
 });
 
-const iterativeCmlRunnerTpl = (opts = {}) => ({
-  ...iterativeProviderTpl(opts),
-  resource: {
-    iterative_cml_runner: {
-      runner: {
-        ...(opts.awsSecurityGroup && {
-          aws_security_group: opts.awsSecurityGroup
-        }),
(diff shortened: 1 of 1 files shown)


0x2b3bfa0:
Was the `tfObj` assignment necessary? 😅

dacbd:
> Was the `tfObj` assignment necessary? 😅

makes hacking and future debugging easier.

0x2b3bfa0:
I'd prefer proper unit testing over `sed` hacking, and confess being allergic to unnecessary variables. 🤧 Still, no hard opinions on this.

dacbd:
@0x2b3bfa0 do you have strong feelings about [`DEBUG`](https://github.com/winstonjs/winston#logging-levels) level logging it, IMO this is important information that I often go to for diagnosing issues that it is overly annoying to get at otherwise...

0x2b3bfa0:
@dacbd, if it's any better than a mere `cat ~/.cml/*/main.tf.json` after the command, sounds fine to me.

dacbd:
> cat ~/.cml/*/main.tf.json

considering you have to nullglob and possibly get a large number of results or annoying extract the `cml-*` id out of the output yes I would prefer a zero effort guaranteed print of the information that I want ⏯️

0x2b3bfa0 (approved):

PR #1133, opened 2022-08-19
Update `yargs-parser`

0x2b3bfa0:
Closes https://github.com/iterative/cml/issues/852 🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7588,8 +7588,9 @@
       }
     },
     "node_modules/yargs-parser": {
-      "version": "21.0.1",
-      "license": "ISC",
+      "version": "21.1.1",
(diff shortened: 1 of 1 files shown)


dacbd (changes requested):
Im assuming this is supposed to allow: `--cloud-ssh-private="$(cat ./key.pem)"` instead of `--cloud-ssh-private="$(cat ./key.pem) "` which doesnt appear to be the case.

0x2b3bfa0:
@dacbd, sounds like [a stray space character](https://github.com/iterative/cml-playground/actions/runs/2890590522/workflow#L79) in your test workflow:

```diff
- --labels=private-key-852 -test 
+ --labels=private-key-852-test 
```

dacbd:
indeed

dacbd (approved):
test fixed, and passes

PR #1166, opened 2022-09-13
Add support for `cml comment <verb> --publish-url`

0x2b3bfa0:
Required for https://github.com/iterative/cml.dev/issues/316
```
cml comment <create|update> --publish --publish-url=https://{identifier}.lambda-url.{region}.on.aws
```

### DIFF
--- a/src/cml.js
+++ b/src/cml.js
@@ -167,6 +167,7 @@ class CML {
       update,
       pr,
       publish,
+      publishUrl,
       markdownFile,
       report: testReport,
       watch,
@@ -208,7 +209,11 @@ class CML {
           );
           if (!triggerFile && watch) watcher.add(absolutePath);
           try {
-            node.url = await this.publish({ ...opts, path: absolutePath });
+            node.url = await this.publish({
+              ...opts,
+              path: absolutePath,
+              url: publishUrl
+            });
           } catch (err) {
             if (err.code !== 'ENOENT') throw err;
           }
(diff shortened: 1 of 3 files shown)


src/cml.js
```diff
@@ -208,7 +209,11 @@ class CML {
           );
           if (!triggerFile && watch) watcher.add(absolutePath);
           try {
-            node.url = await this.publish({ ...opts, path: absolutePath });
+            node.url = await this.publish({
```
tasdomas, line 212: Should we have  a test for this?

tasdomas (approved):
LGTM, but we may want to have a test for this option.

src/cml.js
0x2b3bfa0: Would https://github.com/iterative/cml/pull/1166/commits/75bb629362c46765cd4d37e1a4cef817460229e2 be enough?

tasdomas: won't we be uploading artifacts for every test run?

0x2b3bfa0: *Cough,* yes, in negligible proportion when compared to the legacy `cml publish` tests. 😅 

https://github.com/iterative/cml/blob/dc6749c3fb79a5052cf3ebc3d77a9f4b940da7cd/bin/cml/asset/publish.test.js#L33-L109

tasdomas (approved):

dacbd:
@0x2b3bfa0 I haven't tried this yet, know if you can save me an iteration; does the `cml comment` version of this flag respect the `CML_PUBLISH_URL` env that using `cml publish` does?

0x2b3bfa0:
@dacbd, no, it doesn't

0x2b3bfa0:
> _should it?_

No, we don't need any kind of backwards compatibility here

0x2b3bfa0:
> _`CML_COMMENT_PUBLISH_URL` is what yargs would infer?_

Yes

PR #1172, opened 2022-09-20
Enhance telemetry data and exclude `tensorboard` tracebacks

0x2b3bfa0:
* Fix `cml tensorboard` logging whole Python Traceback to the telemetry backend
* Allow to specify which options get logged to the telemetry backend

### Hacks

* https://github.com/iterative/cml/pull/1172#pullrequestreview-1124464226


package.json
```diff
@@ -82,11 +83,13 @@
     "git-url-parse": "^13.1.0",
     "globby": "^11.0.4",
     "https-proxy-agent": "^5.0.1",
+    "is-docker": "^2.2.1",
```
dacbd, line 86: ahh, node ecosystem. Since this package has no other external dependencies let's pin an exact version and let one of the numerous bots notify us of any required updates.

0x2b3bfa0: Is this package less trustworthy than any of the others? 🙃

bin/cml.js
```diff
@@ -76,9 +76,43 @@ const setupLogger = (opts) => {
   });
 };
 
-const setupTelemetry = async (opts) => {
-  const { cml, cmlCommand: action } = opts;
-  opts.telemetryEvent = await jitsuEventPayload({ action, cml });
+const setupTelemetry = async (opts, { parsed: { defaulted } }) => {
```
DavidGOrtega, line 79: Why such a complex signature?

```js
{ parsed: { defaulted } 
```

do we expect `parsed` to be more complex?

bin/cml.js
```diff
@@ -76,9 +76,43 @@ const setupLogger = (opts) => {
...
+            allowedTelemetryOptions[action] || {},
+            key
+          ) &&
+          !Object.prototype.hasOwnProperty.call(camelcaseKeys(defaulted), key)
+      )
+      .map(([key, value]) => [
+        key,
+        (allowedTelemetryOptions[action] || {})[key] === 'plain' ? value : '***'
+      ])
+  );
```
DavidGOrtega, line 109: 🤯 
This could be done much more simpler and clear

bin/cml/tensorboard/connect.js
```diff
@@ -28,8 +29,8 @@ const tbLink = async (opts = {}) => {
     chrono = chrono + chronoStep;
   }
 
-  const error = await fs.readFile(stderror, 'utf8');
-  throw new Error(`Tensorboard took too long. ${error}`);
+  winston.error(await fs.readFile(stderror, 'utf8'));
```
DavidGOrtega, line 32: This is also fixing TB error stacktrace telemetry. An issue that its not opened as an issue, just in discord.
I would either open an issue and close it here or change the PR to something that may include this?

package.json
```diff
@@ -82,11 +83,13 @@
...
     "js-base64": "^3.7.2",
     "kebabcase-keys": "^1.0.0",
     "node-fetch": "^2.6.5",
     "node-ssh": "^12.0.0",
     "os-name": "^5.0.1",
+    "pascalcase-keys": "^1.0.1",
```
DavidGOrtega, line 92: Why do we need this?

package.json
```diff
@@ -73,6 +73,7 @@
     "@octokit/plugin-throttling": "^3.5.2",
     "@octokit/rest": "18.0.0",
     "appdirs": "^1.1.0",
+    "camelcase-keys": "^7.0.2",
```
DavidGOrtega, line 76: were not we using this before? 🤔

DavidGOrtega (changes requested):

package.json
0x2b3bfa0: No

package.json
0x2b3bfa0: A lapsus from a previous experiment, thanks! 🤦🏼‍♂️

package.json
dacbd: no, and that is my point. It has no dependencies and is unlikely to change so pin the exact version...

Dockerfile
```diff
@@ -116,6 +116,7 @@ WORKDIR ${CML_RUNNER_PATH}
 # SET SPECIFIC ENVIRONMENT VARIABLES
 ENV IN_DOCKER=1
 ENV RUNNER_ALLOW_RUNASROOT=1
+ENV _CML_CONTAINER_IMAGE=true
```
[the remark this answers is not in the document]
dacbd: probably, for the yargs parsing of env for flag options

bin/cml.js
dacbd: @0x2b3bfa0 doesn't believe in comments :joy:

0x2b3bfa0: `// It does what you think it does`

package.json
0x2b3bfa0: It's unlikely to change but already on version 2.2.1 🤔

0x2b3bfa0: Software is never finished, just abandoned

dacbd: as a security-minded person, I'm surprised you don't see the point I'm making as with other cases of other random packages, but whatever...

0x2b3bfa0: I pride myself of seeing the point: having a watering hole less in the project.

My hesitation is still based on the fact that this is just a very small portion of our dependency tree and equally or less likely to suffer a supply chain attack than any of the other dependencies.

Lock files are meant to pin dependencies as you intend. Unless my Node.js notions are wrong (thing I don't discard), the `package.json` file has another purpose.

Dockerfile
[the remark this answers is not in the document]
0x2b3bfa0: _Richtig!_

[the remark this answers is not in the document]
0x2b3bfa0: Otherwise, it would colide with the `CML` namespace and cause the program to fail.

bin/cml.js
0x2b3bfa0: How would you write it? Just take `parsed` or the whole `yargs` and destructure `defaulted` in the function body?

bin/cml.js
0x2b3bfa0: > _This could be done much more simpler and clear_

Ahem, ahem: https://github.blog/2019-02-14-introducing-draft-pull-requests

0x2b3bfa0: And note that `Object.prototype.hasOwnProperty.call(object, property)` was a linter whim; I would have used `object.hasOwnProperty(property)`

package.json
dacbd: I've seen our lock file be rebuilt multiple times from scratch which is based on the package.json file, so all it would take is some unfortunate timing.

I agree it is a small proportion so why should we even allow the possibility when there is nothing to be gained?[^1]

[^1]: nothing wrong with that; perhaps suggest to him that as a very prolific open source developer he sign commits his commits and tags?

bin/cml/tensorboard/connect.js
0x2b3bfa0: Updated the pull request body

package.json
0x2b3bfa0: > _I've seen our lock file be rebuilt multiple times from scratch which is based on the package.json file, so all it would take is some unfortunate timing._

I'd argue that it's partly because of some development “practices” like running `npm install` instead of `npm ci` and committing unsupervised lock file changes on every single pull request; and partly because the Node.js package ecosystem is rather “optimistic” to put it mildly.

> _I agree it is a small proportion so why should we even allow the possibility when there is nothing to be gained?_

We'd gain consistency. 🤷🏼‍♂️

0x2b3bfa0: Anyway, resolved with 4880fd7

0x2b3bfa0: Should we ask #javascript for more opinions on the topic? It would be nice to do this always if we find it's a good idea, or never if we don't.

bin/cml/check/create.js
```diff
@@ -10,7 +10,11 @@ exports.handler = async (opts) => {
   await cml.checkCreate({ ...opts, report });
 };
 
-exports.builder = (yargs) => yargs.env('CML_CHECK').options(exports.options);
+exports.builder = (yargs) =>
+  yargs
+    .env('CML_CHECK')
+    .option('options', { default: exports.options, hidden: true })
```
0x2b3bfa0, line 16: This is a gross hack, because otherwise there is no way[^1] of accessing the options object from a middleware.

[^1]: None that I know of

dacbd (changes requested):
something is not right: https://github.com/iterative/cml-playground/actions/runs/3147876322/jobs/5117818474

dacbd (dismissed):

0x2b3bfa0:
@DavidGOrtega?

Dockerfile
tasdomas, line 119:
```suggestion
# Environment variable used by cml to detect it's been installed using the docker image.
ENV _CML_CONTAINER_IMAGE=true
```

tasdomas (approved):

tasdomas (dismissed):

.github/workflows/release.yml
```diff
@@ -31,11 +31,9 @@ jobs:
...
       - uses: actions/checkout@v3
       - run: >
-          gh release create
-          --target ${{ github.event.pull_request.merge_commit_sha }}
-          {--title=CML\ ,}$(basename ${{ github.head_ref }})
-          --generate-notes
-          --draft
+          gh release create --target ${{
+          github.event.pull_request.merge_commit_sha }} {--title=CML\
+          ,}$(basename ${{ github.head_ref }}) --generate-notes --draft
```
0x2b3bfa0, line 36:
```suggestion
          gh release create
          --target ${{ github.event.pull_request.merge_commit_sha }}
          {--title=CML\ ,}$(basename ${{ github.head_ref }})
          --generate-notes
          --draft
```

.github/workflows/images.yml
```diff
@@ -94,8 +94,9 @@ jobs:
       - uses: docker/build-push-action@v2
         with:
           push:
-            ${{ inputs.release || github.event_name == 'push' || github.event_name == 'schedule'
-            || github.event_name == 'workflow_dispatch' }}
+            ${{ inputs.release || github.event_name == 'push' ||
+            github.event_name == 'schedule' || github.event_name ==
+            'workflow_dispatch' }}
```
0x2b3bfa0, line 99:
```suggestion
            ${{ inputs.release || github.event_name == 'push' || github.event_name == 'schedule'
            || github.event_name == 'workflow_dispatch' }}
```

dacbd (approved):

PR #1173, opened 2022-09-20
Augment watermark with label

### LINKED ISSUE #1144: custom watermark ID
So that when cml goes to update a comment it can determine which comment to update.

For cases where multiple comments are made or when multiple workflows run against a single PR each with a cml comment, they can better distinguish which comment to update.

tasdomas:
This will allow us to resolve updatable comments to a user-specified scope, instead of updating the most recent comment containing the string `watermark.svg`, as we currently do.

## commentLabel
The `--commentLabel` flag allows users to specify comment labels of any type and specificity. For ease of use, special placeholder strings `{workflow}` and `{run}` which will be replaced with the workflow id or id of the current execution.

The resulting watermark will be of the form 
```
![](https://cml.dev/watermark.png "CML watermark <user-specified label>")
```

Obviously this PR will only be merged once studio has been updated to detect the new version of CML watermarks.

This closes #1144


src/cml.js
```diff
@@ -159,19 +159,40 @@ class CML {
     throw new Error(`driver ${driver} unknown!`);
   }
 
+  watermark(driver, scope = 'global') {
+    var tag;
+    var id;
+    switch (scope) {
+      case 'global':
+        return '![CML watermark](https://raw.githubusercontent.com/iterative/cml/master/assets/watermark.svg)';
+      case 'workflow':
```
dacbd, line 168: I'd have to revisit my thoughts on this, but I believe this should really be a combo of both these options. I can have two files with the same job id.

or perhaps another scope as an option that tries to be more unique or a user-provided tag.

dacbd: jobId() is reading from `GITHUB_JOB` which is user defined. you can have two separate workflows with the same job id. 

I can have:
```yaml
name: thing1
on: ...
jobs:
  train:
```
and
```yaml
name: thing2
on: ...
jobs:
  train:
```

where id will be `train` for both workflows. thus it should either be `workflow-id + job-id` to be unique or this in addition to what is present. note that workfowId() would return `thing1` and `thing2` respectively here and jobId() would return `train` for both.

tasdomas: `job` was a choice for scope name on my part - I've changed it to `run` (see updated PR comment)

DavidGOrtega: @tasdomas this change might break Studio. Let me check the studio code.

DavidGOrtega (changes requested):

tasdomas:
@DavidGOrtega this change will definitely break studio, since it currently [searches comment bodies](https://github.com/iterative/studio/blob/develop/backend/repos/tasks/cml.py#L153) for the exact markdown representation of the CML watermark.

But I think the watermark format needs revising and a decision on both how we augment the watermark format and how we detect its presence in comments (cml itself, for example, [only searches](https://github.com/iterative/cml/blob/master/src/cml.js#L266) for the presence of the string `watermark.svg` to detect cml-marked comments).

DavidGOrtega:
> @DavidGOrtega this change will definitely break studio, since it currently [searches comment bodies](https://github.com/iterative/studio/blob/develop/backend/repos/tasks/cml.py#L153) for the exact markdown representation of the CML watermark.
> 
> But I think the watermark format needs revising and a decision on both how we augment the watermark format and how we detect its presence in comments (cml itself, for example, [only searches](https://github.com/iterative/cml/blob/master/src/cml.js#L266) for the presence of the string `watermark.svg` to detect cml-marked comments).

So this PR should be blocked until we add code in Studio to fix it. 
Opens https://github.com/iterative/studio/issues/4389

DavidGOrtega:
> @DavidGOrtega this change will definitely break studio, since it currently [searches comment bodies]

You might want to attack it? I can follow up with you on "put your hands on Studio" @tasdomas

tasdomas:
> > @DavidGOrtega this change will definitely break studio, since it currently [searches comment bodies]
> 
> I have taken it but you might want to attack it. I can follow up with you on "put your hands on Studio" @tasdomas

Thanks, @DavidGOrtega - once we've agreed on a watermark format, I'd be happy to update studio as well

src/utils.js
```diff
@@ -136,14 +136,14 @@ const isProcRunning = async (opts) => {
...
+  return uriParam({ uri, param: 'cml', value: type });
 };
 
 const preventcacheUri = ({ uri } = {}) => {
-  return uriParmam({ uri, param: 'cache-bypass', value: uuid.v4() });
+  return uriParam({ uri, param: 'cache-bypass', value: uuid.v4() });
 };
 
-const uriParmam = (opts = {}) => {
+const uriParam = (opts = {}) => {
```
tasdomas, line 146: This is a simple drive-by fix.

bin/cml/comment/create.js
```diff
@@ -55,5 +55,12 @@ exports.options = kebabcaseKeys({
     type: 'boolean',
     description:
       'Avoid watermark; CML needs a watermark to be able to distinguish CML comments from others'
+  },
+  commentLabel: {
```
casperdcl, line 59: to avoid duplication (`cml comment ... --comment`)

```suggestion
  id: {
```

src/drivers/bitbucket_cloud.js
```diff
@@ -421,6 +425,14 @@ class BitbucketCloud {
     return command;
   }
 
+  get workflowId() {
+    return BITBUCKET_PIPELINE_UUID;
```
casperdcl, line 429: doesn't this change for each run?

bin/cml/comment/create.js
```diff
@@ -55,5 +55,12 @@ exports.options = kebabcaseKeys({
...
+    type: 'string',
+    description:
+      'Label to add to watermarks for more selective comment updates',
+    default: '',
```
casperdcl, line 63: does it make sense to do this by default?

```suggestion
    default: '{workflow}',
```

bin/cml/comment/create.js
```diff
@@ -55,5 +55,12 @@ exports.options = kebabcaseKeys({
     type: 'boolean',
     description:
       'Avoid watermark; CML needs a watermark to be able to distinguish CML comments from others'
+  },
+  commentLabel: {
+    type: 'string',
+    description:
+      'Label to add to watermarks for more selective comment updates',
```
casperdcl, line 62: mention `{workflow,run}` template vars?

```suggestion
      'Comment identifier, where {workflow} and {run} are auto-replaced',
```

casperdcl (changes requested):
need to mod `comment/update.*js` too

src/drivers/bitbucket_cloud.js
tasdomas: Yes. And that's the point - while the user may want to update a comment as the run progresses, they may want to keep comments from previous runs unmodified.

bin/cml/comment/create.js
tasdomas: That would alter the current behavior.

bin/cml/comment/create.js
tasdomas: how about `label`?

tasdomas:
> need to mod `comment/update.*js` too

`comment/update.js` does not define additional flags

dacbd (approved):
LGTM: https://github.com/iterative/cml-playground/pull/109
https://github.com/iterative/cml-playground/blob/main/.github/workflows/cml-1173.yml

bin/cml/comment/create.js
casperdcl: unfortunately that overloads/conflicts with:

- [CML runner labels](https://cml.dev/doc/ref/runner#--labels)
- [GH labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels)

bin/cml/comment/create.js
casperdcl: yes, it would - but is that a gewd thing?

src/drivers/bitbucket_cloud.js
dacbd: I thought the pipeline uuid stayed the same?

bin/cml/comment/create.js
tasdomas: `tag`?

src/drivers/bitbucket_cloud.js
dacbd: @casperdcl I think this is safe to resolve, using `BITBUCKET_PIPELINE_UUID` is functionally eq to GH

bin/cml/comment/create.js
dacbd: at the risk of being to verbose, I would learn towards `watermarkId` or `watermarkTag`. Certainly not plain `id` though?

tasdomas: `watermark` is arguably an internal cml concept, that should have limited exposure to users - `commentTag`?

dacbd: `commentTag` 👍 lets go with that.

bin/cml/comment/create.js
tasdomas: I've changed the url of the image used in the watermark - it's now served from cml.dev

src/cml.js
```diff
@@ -159,19 +160,33 @@ class CML {
     throw new Error(`driver ${driver} unknown!`);
   }
 
+  watermark(driver, label = '') {
+    if (label === '') {
+      return `![](${WATERMARK_IMAGE} "CML watermark")`;
+    }
```
casperdcl, line 166: minor, but do we prefer this for readability?

```suggestion
  watermark(driver, label = 'CML watermark') {
```

bin/cml/comment/create.js
casperdcl: not a fan (clashes with Git tags).

`--id` seems best UX to me. It doesn't clash with anything, at least as far as end-users are concerned.

src/cml.js
tasdomas: That's not exactly the same, though

src/cml.js
```diff
@@ -159,19 +160,33 @@ class CML {
...
+    // Replace {workflow} and {run} placeholders in label with actual values.
+    const workflow = driver.workflowId;
+    const run = driver.runId;
+    label = label.replace('{workflow}', workflow);
+    label = label.replace('{run}', run);
+    label = `CML watermark ${label}`;
+    return `![](${WATERMARK_IMAGE} "${label}")`;
+  }
```
0x2b3bfa0, line 174:
```suggestion
  watermark(driver, label = '') {
    // Replace {workflow} and {run} placeholders in label with actual values.
    const workflow = driver.workflowId;
    const run = driver.runId;
    label = label.replace('{workflow}', workflow);
    label = label.replace('{run}', run);
    label = `CML watermark ${label}`.trim();
    return `![](${WATERMARK_IMAGE} "${label}")`;
  }
```

src/cml.js
0x2b3bfa0: https://github.com/iterative/cml/pull/1173#discussion_r990919162 for identical behavior? 😸

bin/cml/comment/create.js
tasdomas: Let's take a step back and see which command line makes more sense:
```
cml comment create --update --pr --publish --comment-label='{run}-best-result'
```
or 
```
cml comment create --update --pr --publish --id='{run}-best-result'
```

dacbd: the group has decided on `--cml-comment-id`

bin/cml/comment/create.js
```diff
@@ -63,5 +63,12 @@ exports.options = kebabcaseKeys({
     description:
       'Avoid watermark; CML needs a watermark to be able to distinguish CML comments from others',
     telemetryData: 'name'
+  },
+  watermarkTitle: {
+    type: 'string',
+    description:
+      'Watermark title to add to comments for selective updates, {workflow} and {run} are auto-replaced',
```
casperdcl, line 70:
```suggestion
      'Hidden comment marker (used for targeting in subsequent `cml comment update`); "{workflow}" & "{run}" are auto-replaced',
```

src/cml.js
```diff
@@ -160,19 +161,30 @@ class CML {
...
 
+  watermark(driver, label = '') {
+    // Replace {workflow} and {run} placeholders in label with actual values.
+    const workflow = driver.workflowId;
+    const run = driver.runId;
+    label = label.replace('{workflow}', workflow);
+    label = label.replace('{run}', run);
+    label = `CML watermark ${label}`.trim();
+    return `![](${WATERMARK_IMAGE} "${label}")`;
+  }
```
DavidGOrtega, line 172: Makes no a big sense to make this function as a member of the class and pass the driver. It can retrieve the driver at any point. Also the signature of the functions is always passing an object as argument vs arguments. This way we then do named parameters calls.

I suggest to remove this function and make it within the specific function

```suggestion
```

src/cml.js
```diff
@@ -160,19 +161,30 @@ class CML {
...
       publishUrl,
-      markdownFile,
       report: testReport,
+      rmWatermark,
+      triggerFile,
+      update,
       watch,
-      triggerFile
+      watermarkTitle
     } = opts;
```
DavidGOrtega, line 188:
```suggestion
      watermarkTitle
    } = opts;
    
    const drv = this.getDriver();
```

src/cml.js
```diff
@@ -181,9 +193,8 @@ class CML {
     if (rmWatermark && update)
       throw new Error('watermarks are mandatory for updateable comments');
 
-    const watermark = rmWatermark
-      ? ''
-      : '![CML watermark](https://raw.githubusercontent.com/iterative/cml/master/assets/watermark.svg)';
+    const drv = this.getDriver();
+    const watermark = rmWatermark ? '' : this.watermark(drv, watermarkTitle);
```
DavidGOrtega, line 197:
```suggestion
const genWatermark = () => {
    if (rmWatermark) return '';
    const { workflowId, runId } = drv;
    const lbl = watermarkTitle.replace('{workflow}', workflowId).replace('{run}', runId).trim();
    return `![](${WATERMARK_IMAGE} "CML watermark ${lbl}")`;
  }
  const watermark = genWatermark();
```

DavidGOrtega (changes requested):
🙏 some considerations

src/cml.js
tasdomas: I understand adhering to a single function signature across the codebase. But I don't see how it's logical to merge the code for watermark creation into the `commentCreate` function's body. That function (`commentCreate`) is already over 150 lines long, my instinct is to remove code from it, not add more.

DavidGOrtega: I saw it, I would suggest a refactor later on replacing section of code within functions.
That way despite the method is still large it can be collapsed logically.
Reasons to not do the watermark method:

 - There is not point since its not very useful interfacing outside CML. The method in js is public (not private) and mangles a bit the interface
 
In any case the strangest thing is the driver passed as a parameter. It can get it at any point.
You can externalise the function also within the module like many others

```js
const genWatermark = ({ runId, workflowId, rmWatermark, watermarkTitle } = {}) => {
    if (rmWatermark) return '';
    const lbl = watermarkTitle.replace('{workflow}', workflowId).replace('{run}', runId).trim();
    return `![](${WATERMARK_IMAGE} "CML watermark ${lbl}")`;
  }
```

then call genWatermark

DavidGOrtega: In any case if we were encapsulating code within functions inside the member would be fine and should not be counted as X lines of code

src/cml.js
DavidGOrtega: I have opened #1234 @tasdomas

src/cml.js
DavidGOrtega: I have opened https://github.com/iterative/cml/issues/1234 @tasdomas

tasdomas: Sure, I'll refactor.

But not counting embedded functions towards the line count seems like a bad way to move towards a readable and understandable codebase.

DavidGOrtega: lets see in the refactor... I have also the feeling that we are making it a bit complex in general

src/cml.js
```diff
@@ -164,26 +165,43 @@ class CML {
...
+    // Create the watermark.
+    const genWatermark = ({ label, workflow, run } = opts) => {
+      // Replace {workflow} and {run} placeholders in label with actual values.
+      label = label
+        .replace('{workflow}', workflow)
+        .replace('{run}', run)
+        .trim();
+      const title = `CML watermark ${label}`.trim();
+      return `![](${WATERMARK_IMAGE} "${title}")`;
+    };
```
DavidGOrtega, line 197:
```suggestion
    const genWatermark = ({ label, workflow, run } = opts) => {
      const title = label
        .replace('{workflow}', workflow)
        .replace('{run}', run)
        .trim();
      return `![](${WATERMARK_IMAGE} "CML watermark ${title}")`;
    };
```

DavidGOrtega: I think the comment do not add a lot of value... can be seen fast in the code

Also I prefer to not modify passed values. In this case works but js has funny things like [call-by-sharing](http://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_sharing) and its safer to just consider the pass as reference

I think that you do not need to trim twice 🙂

DavidGOrtega (changes requested):
🙏 some minor

src/cml.js
DavidGOrtega: Its also a bit strange label vs title

tasdomas: We need the second trim, so that if `label` is empty, the resulting watermark will be ``, not `` (the extra space at the end of the watermark title)

DavidGOrtega: Ok you do not need the first trim then

tasdomas:
@DavidGOrtega - does it look better to you now?

src/cml.js
tasdomas: dropped it - thanks!

DavidGOrtega (dismissed):
lgtm 🚀

dacbd (changes requested):
Seems that a bug was introduced or the behavior has changed since my last review.

Lastest test run: https://github.com/iterative/cml-playground/pull/135

should match the behavior seen here: https://github.com/iterative/cml-playground/pull/109

tasdomas:
> Seems that a bug was introduced or the behavior has changed since my last review.

took me a while to track down - apparently something is escaping `custom_2` into `custom\_2` (underscores preceding numerals are escaped). It looks like it's `octokit`. Working on figuring out the escape rules.

tasdomas:
> Seems that a bug was introduced or the behavior has changed since my last review.
> 
> Lastest test run: [iterative/cml-playground#135](https://github.com/iterative/cml-playground/pull/135)
> 
> should match the behavior seen here: [iterative/cml-playground#109](https://github.com/iterative/cml-playground/pull/109)

I reran the jobs on [the cml-playground PR](https://github.com/iterative/cml-playground/pull/135) - looks like [1400685](https://github.com/iterative/cml/pull/1173/commits/14006850423ce15d69c3c4c5b1206311c7f15013) fixes the issue.

dacbd:
@tasdomas It does not look that worked properly, comments with `cml comment create --pr --watermark-title "{workflow}" ./comment.md` were not updated. https://github.com/iterative/cml-playground/blob/8fbaa9bcf989e783c9bcdc026c7dd9beb772e77f/.github/workflows/cml-1173.yml#L55-L78 and https://github.com/iterative/cml-playground/blob/main/.github/workflows/cml-1173-2.yml have never been edited even after push additional commits to re-trigger the run. The two workflow_label ones should have been edited.

tasdomas:
> @tasdomas It does not look that worked properly, comments with `cml comment create --pr --watermark-title "{workflow}" ./comment.md` were not updated. https://github.com/iterative/cml-playground/blob/8fbaa9bcf989e783c9bcdc026c7dd9beb772e77f/.github/workflows/cml-1173.yml#L55-L78 and https://github.com/iterative/cml-playground/blob/main/.github/workflows/cml-1173-2.yml have never been edited even after push additional commits to re-trigger the run. The two workflow_label ones should have been edited.

@dacbd - looks like I finally got it.

dacbd (dismissed):

tasdomas:
> @DavidGOrtega / @tasdomas the required studio change has already been made, correct?

yes, not sure if it's been released yet

DavidGOrtega:
Should be available since 1.20.0

dacbd (approved):

PR #1185, opened 2022-09-26
Fix failing gitlab workflow.

tasdomas:
Defining port mappings without quotes is ambiguous yaml syntax.


DavidGOrtega (approved):
🤯 

Expedited approval

PR #1208, opened 2022-10-06
Enable `cml comment create --publish` by default

0x2b3bfa0:
Behavior only changes for local Markdown links


tasdomas (approved):

dacbd:
do we only publish if the file isn't in git, ie you can reference local files that are in your git repo in markdown for links too and they will be rendered(without using `--publish`) right?

0x2b3bfa0:
After this change, any local file referenced from the Markdown document will be automatically uploaded

PR #1238, opened 2022-10-26
Configure winston logging when running terraform tests

tasdomas:
Otherwise warnings are printed about unconfigured transports.


0x2b3bfa0 (approved):
Thanks!

PR #1257, opened 2022-11-15
Add a .nvmrc file to support using the intended Node.js version

magdapoppins:
(Dev experience suggestion - feel free to close the pr if this doesn't seem like a good addition.)

As per https://cml.dev/doc/contributing/core, 16.x is the correct Node version to use for this repo - adding an .nvmrc file allows contributors that have nvm at hand to simply call `nvm use` to autoselect the projects intended Node version before `npm install`.


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
dacbd, line 1:
```suggestion
16

```

dacbd (approved):
added newline to end of file for my OCD

casperdcl (commented):
but, but, but... https://github.com/iterative/cml/blob/65a200aa0042d8ce219933346741f9fb2348995f/package.json#L33

(my OCD hates dotfile root clutter)

dacbd:
ㅋㅋㅋ we can revert if your OCD becomes too much ❤️

If it encourages another open source contributor my vote is +1.

PR #1262, opened 2022-11-30
Make comment targets case-insensitive.

tasdomas:
Added this, because for some reason my lizard brain found typing `--target=PR#123` more natural. Either way, I don't think having this will hurt.


0x2b3bfa0 (approved):
Sounds useful, although normalization on behalf of users is a whole field of study

PR #1265, opened 2022-12-01
Default cml name change paired with (tpi#726)

dacbd:


0x2b3bfa0 (commented):
I was going to do exactly this yesterday, but then noticed that `--name` is being used for both the cloud resource names and to register the self-hosted runner on GitHub (read the option description)

I wonder if we should keep the `cml` prefix on the latter.

dacbd:
hmm, I don't have strong opinions on this, and I don't think that many people look at that view in GitHub (list of all of your self-hosted actions runners) but they definitely look at their web console. (the names are referred to in a few other places)

@0x2b3bfa0 I think we should keep it simple, and not change that yet. If someone looks at the screen enough they could use their own name generation that makes sense to them. We can cross these bridges one at a time easily enough.

0x2b3bfa0 (approved):
🤷🏼‍♂️

0x2b3bfa0:
Container builds are broken, but not because of this pull request

PR #1286, opened 2022-12-15
Skip failing github runner token e2e test.

tasdomas:
The test is failing due to missing permissions in the test token we use for our integration tests. The
skipped test should be reenabled once that is fixed.


dacbd:
I put the old token back in place and triggered a rerun of tests for the open PRs (something is more broken than I thought with the new style PAT)

PR #1304, opened 2023-01-03
Publish includes sha hash

dacbd:
(https://iterativeai.slack.com/archives/C01116VLT7D/p1671751623131269)

### DIFF
--- a/src/cml.js
+++ b/src/cml.js
@@ -300,8 +300,6 @@ class CML {
     const updatableComment = (comments) => {
       return comments.reverse().find(({ body }) => {
         // break body to find and strip watermark
-        console.log(body);
-        console.log(body.split('\n'));
         return body
           .split('\n')
           .map((line) => this.stripWatermarkFragment(line))
--- a/src/cml.js
+++ b/src/cml.js
@@ -184,6 +184,7 @@ class CML {
     if (rmWatermark && update)
       throw new Error('watermarks are mandatory for updateable comments');
 
+    const sha = await this.triggerSha();
     // Create the watermark.
     const genWatermark = (opts = {}) => {
       const { label = '', workflow, run } = opts;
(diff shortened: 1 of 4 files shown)


0x2b3bfa0 (approved):
Looks good to me, but it's still pending resolution of https://github.com/iterative/studio/pull/4808#discussion_r1061251480 to determine whether to use the fragment or the query to store commit hashes.

casperdcl (commented):
erm... don't we need to update `genWatermark` (https://github.com/iterative/cml/blob/506e4e8d60e48e2ee4e99cfe9ff38ad6ec6a8f7f/src/cml.js#L206) instead?

src/cml.js
```diff
@@ -339,7 +340,9 @@ class CML {
 
     if (!rmWatermark) {
       const [, type] = mime.split('/');
+      const sha = await this.triggerSha();
       uri = watermarkUri({ uri, type });
+      uri = addshaUri({ uri, sha });
```
tasdomas, line 345: I don't think this is the correct place to add this. This just adds a query parameter to urls of published resources.

tasdomas (changes requested):
I don't think this will solve our issue with the studio integration.

We need to augment `WATERMARK_IMAGE` with a query fragment indicating the commit sha in [`genWatermark`](https://github.com/iterative/cml/pull/1304/files#diff-509933ea8efc212e9e460c2b844fcbd03061d80cef264781b632118f971e4584L187), but then the tricky bit is updating the logic we use to determine updatable comments [here](https://github.com/iterative/cml/pull/1304/files#diff-509933ea8efc212e9e460c2b844fcbd03061d80cef264781b632118f971e4584L295) and [here](https://github.com/iterative/cml/pull/1304/files#diff-509933ea8efc212e9e460c2b844fcbd03061d80cef264781b632118f971e4584L310)

dacbd:
I tested here to verify the changes work, but feel free to re-test in another way. https://github.com/iterative/cml-playground/pull/286

src/cml.js
tasdomas: Do we need this as well?

PR #1309, opened 2023-01-11
Add the sha of the the current commit to the watermark as the fragment of the watermark image url.

tasdomas:
This PR also updates the way updatable comments are determined - the matcher should ignore the sha in
the watermark.

This PR supercedes #1304.

### DIFF
--- a/src/cml.js
+++ b/src/cml.js
@@ -276,7 +276,7 @@ class CML {
     let comment;
     const updatableComment = (comments) => {
       return comments.reverse().find(({ body }) => {
-        return watermark === null || watermark.isIn(body);
+        return !watermark || watermark.isIn(body);
       });
     };
 
--- a/src/cml.js
+++ b/src/cml.js
@@ -189,7 +189,7 @@ class CML {
           label: watermarkTitle,
           workflow: drv.workflowId,
           run: drv.runId,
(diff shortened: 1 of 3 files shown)


src/cml.js
```diff
@@ -294,7 +276,7 @@ class CML {
     let comment;
     const updatableComment = (comments) => {
       return comments.reverse().find(({ body }) => {
-        return body.includes(watermark);
+        return watermark === null || watermark.isIn(body);
```
DavidGOrtega, line 279: silly nitpick just in case is `undefined`
```suggestion
        return !watermark || watermark.isIn(body);
```

DavidGOrtega (changes requested):
Just only a little observation that might be even incorrect

src/cml.js
tasdomas: That's a good catch - thank you!

tasdomas: Addressed in https://github.com/iterative/cml/pull/1309/commits/84403766e70647da2175ce93af6629d1df322ccc

dacbd (approved):
I like it more than mine.
https://github.com/iterative/cml-playground/pull/294

DavidGOrtega (approved):
lgtm 🚀

PR #1311, opened 2023-01-12
Stop pushing to docker.io/dvcorg

0x2b3bfa0:
Deprecated as per an internal notice; we'll see if any users still rely on it by means of [fault-induced telemetry](https://github.com/iterative/cml/pull/1003#issuecomment-1124246143). 😸


0x2b3bfa0:
Not needed

Issue #1348, opened 2023-02-10
Lots of images fail to load in one report


shortcipher3:
Running some `cml comment` with a report containing maybe 10 images that I wanted displayed, it is cutting off the last two (it varies a little, sometimes just one, sometimes more), when I look at the markdown for the images that failed to display I see that the images that failed to upload have the original name, rather than the updated url.

After breaking up the report into multiple comments each with fewer images I was able to get all 10 comments to load.

I'm guessing there is a limit, it might be nice to call that out as a warning or error - if you can help users catch that early.

dacbd:
@shortcipher3 thanks for making an issue, I want to confirm that you were using the GitLab only `--publish-native` flag correct?

I believe you are likely correct and that the comment reached some kind of file size limit due to `--publish-native`

shortcipher3:
Yes, that is correct, I was using gitlab with the `--publish-native` flag.

PR #1350, opened 2023-02-12
GitLab  - check comment body size

dacbd:
https://docs.gitlab.com/ee/api/notes.html#create-new-merge-request-note

Closes: https://github.com/iterative/cml/issues/1348

### DIFF
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -16,7 +16,7 @@ const { CI_JOB_ID, CI_PIPELINE_ID, IN_DOCKER } = process.env;
 const API_VER = 'v4';
 const MAX_COMMENT_SIZE = 1000000;
 const ERROR_COMMENT_SIZE =
-  'GitLab Comment is too large, this is likely caused by the `--publish-native` flag';
+  'GitLab Comment is too large, this is likely caused by the `--publish-native` flag causing the comment to pass the 1M character limit';
 
 class Gitlab {
   constructor(opts = {}) {
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -15,6 +15,9 @@ const { CI_JOB_ID, CI_PIPELINE_ID, IN_DOCKER } = process.env;
 
 const API_VER = 'v4';
 const MAX_COMMENT_SIZE = 1000000;
+const ERROR_COMMENT_SIZE =
+  'GitLab Comment is too large, this is likely caused by the `--publish-native` flag';
+
 class Gitlab {
   constructor(opts = {}) {
     const { repo, token } = opts;
@@ -70,8 +73,7 @@ class Gitlab {
   async commitCommentCreate(opts = {}) {
     const { commitSha, report } = opts;
 
-    if (report.length >= MAX_COMMENT_SIZE)
-      throw new Error('GitLab Comment too Large');
+    if (report.length >= MAX_COMMENT_SIZE) throw new Error(ERROR_COMMENT_SIZE);
 
     const projectPath = await this.projectPath();
     const endpoint = `/projects/${projectPath}/repository/commits/${commitSha}/comments`;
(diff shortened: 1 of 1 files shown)


src/drivers/gitlab.js
```diff
@@ -334,6 +334,8 @@ class Gitlab {
     const projectPath = await this.projectPath();
     const { issueId, report, id: commentId } = opts;
 
+    if (report.length >= 1000000) throw new Error('GitLab Comment too Large');
```
balazspeczeli, line 337: How about creating a constant for this value (to avoid having a magic number)? Something like `MAX_COMMENT_LENGTH`?

We can put it on line 17, just before the class declaration:

```javascript
const API_VER = 'v4';
const MAX_COMMENT_LENGTH = 1000000;

class Gitlab {
```

dacbd:
Makes sense to me, I'll push a change soon. This was a pretty quick/dirty
fix.

src/drivers/gitlab.js
```diff
@@ -69,6 +70,9 @@ class Gitlab {
   async commitCommentCreate(opts = {}) {
     const { commitSha, report } = opts;
 
+    if (report.length >= MAX_COMMENT_SIZE)
+      throw new Error('GitLab Comment too Large');
```
casperdcl, line 74: should we warn instead & truncate the report?

incidentally we should also have a generic `warn` function that does the appropriate CI-dependent thing (e.g. on GH, https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message)

tasdomas: https://github.com/iterative/cml/blob/master/src/drivers/github.js#L781

src/drivers/gitlab.js
```diff
@@ -14,6 +14,7 @@ const { fetchUploadData, download, gpuPresent } = require('../utils');
 const { CI_JOB_ID, CI_PIPELINE_ID, IN_DOCKER } = process.env;
 
 const API_VER = 'v4';
+const MAX_COMMENT_SIZE = 1000000;
```
tasdomas, line 17:
```suggestion
const MAX_COMMENT_SIZE = 1024*1024;
```

tasdomas (approved):

src/drivers/gitlab.js
0x2b3bfa0: > [_The content of a note. Limited to 1,000,000 characters._](https://docs.gitlab.com/ee/api/notes.html#create-new-merge-request-note)

I'm afraid it's 1 MB and not 1 MiB

dacbd:
To address the warn vs error. To my knowledge, only GitHub provides a native means to supply these notifications, and this is specific to GitLab.

As far as logging vs an error. The issue the user brought up made it seem like this more or less silently failed, so I think exiting with a non-zero status code is the best option.

How are we going to truncate the report? I don't think we can realistically decide for the user what to drop especially since this is most likely coming from the use of the `--publish-native`. They could end up with a comment that is just one or two lines, if they try and upload a large image.

src/drivers/gitlab.js
casperdcl: can move to follow up.

Also better error message should suggest cause & solutions:

- split into multiple reports
- publish fewer images
- don't use `--publish-native` as it encodes images in `base64`, easily hitting the limit
- what is the max char limit (1M)

casperdcl, line 74: Also better error message should suggest cause & solutions:

- split into multiple reports
- publish fewer images
- don't use `--publish-native` as it encodes images in `base64`, easily hitting the limit
- what is the max char limit (1M)

casperdcl (dismissed):

src/drivers/gitlab.js
```diff
@@ -14,6 +14,10 @@ const { fetchUploadData, download, gpuPresent } = require('../utils');
...
+const ERROR_COMMENT_SIZE =
+  'GitLab Comment is too large, this is likely caused by the `--publish-native` flag';
```
dacbd, line 19:
```suggestion
  'GitLab Comment is too large, this is likely caused by the `--publish-native` flag causing the comment to pass the 1M character limit';
```

0x2b3bfa0 (approved):
Looks good to me

Issue #1365, opened 2023-03-13
Add support for Kubernetes arbitrary node selector


ludelafo:
Hello,

We have an on-premises Kubernetes cluster with multiple nodes. Some of these nodes have dedicated GPUs.

We would like to use CML to train our models on those specific nodes. Kubernetes supports many ways to specify on which node a pod should be run:

- [Assigning Pods to Nodes](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/)
- [Assigning Pods to Nodes #nodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector)
- [Assigning Pods to Nodes #Affinity and anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity)

### nodeSelector

Assigning a pod to a node with `nodeSelector` is as simple as labeling a node with a certain tag and add a `nodeSelector` to the pod configuration. Example inspired by the [Assign Pods to Nodes [with `nodeSelector`]](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/) documentation:

```sh
# Label a node with `disktype=ssd`
kubectl label nodes <your-node-name> disktype=ssd
```

```yaml
# pod-nginx.yaml (note the `nodeSelector`)
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
    - name: nginx
      image: nginx
      imagePullPolicy: IfNotPresent
  nodeSelector:
    disktype: ssd
```

```sh
# Apply the Pod configuration
kubectl apply -f pod-nginx.yaml
```

The pod will be created on the node with the label `disktype=ssd`.

**Bonus**: Assign a pod to a node by its name

You can also assign a pod to a node using the node's name. Example inspired by the [Assign Pods to Nodes [with `nodeSelector`] #Create a pod that gets scheduled to specific node](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/#create-a-pod-that-gets-scheduled-to-specific-node) documentation:

```yaml
# pod-nginx.yaml (note the `nodeName`)
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  nodeName: the-node-name
  containers:
    - name: nginx
      image: nginx
      imagePullPolicy: IfNotPresent
```

### Affinity and anti-affinity

[...]

PR #1370, opened 2023-04-04
Add support for Kubernetes arbitrary node selector

ludelafo:
This PR implements the Kubernetes arbitrary node selector described in issue https://github.com/iterative/cml/issues/1365.

This is a work in progress and my first experience with Golang and Terraform provider development, I'm open to improvements.

Related PRs:

- https://github.com/iterative/terraform-provider-iterative/pull/747
- https://github.com/iterative/cml.dev/pull/464

### DIFF
--- a/bin/cml/runner/launch.js
+++ b/bin/cml/runner/launch.js
@@ -591,18 +591,15 @@ exports.options = kebabcaseKeys({
     alias: 'cloud-aws-subnet-id'
   },
   cloudKubernetesNodeSelector: {
-    type: 'string',
-    default: 'accelerator=inferred',
+    type: 'array',
+    string: true,
+    default: [],
+    coerce: (items) => {
+      const keyValuePairs = items.map((item) => [...item.split(/=(.+)/), null]);
+      return Object.fromEntries(keyValuePairs);
+    },
     description:
-      'Specifies the node selector to use within Kubernetes. By default, the node selector is inferred from the GPU configuration',
-    coerce: (val) => {
-      const [key, value] = val.split(/=(.+)/);
-
-      return {
-        key,
-        value
-      };
-    }
+      'Key Value pairs to specify the node selector to use within Kubernetes i.e. tags/labels "key=value"'
   },
   cloudImage: {
     type: 'string',
--- a/bin/cml/runner/launch.js
+++ b/bin/cml/runner/launch.js
@@ -135,6 +135,7 @@ const runCloud = async (opts) => {
       cloudStartupScript: startupScript,
       cloudAwsSecurityGroup: awsSecurityGroup,
       cloudAwsSubnet: awsSubnet,
+      cloudKubernetesNodeSelector: kubernetesNodeSelector,
       cloudImage: image,
       workdir
     } = opts;
@@ -172,6 +173,7 @@ const runCloud = async (opts) => {
       startupScript,
       awsSecurityGroup,
       awsSubnet,
+      kubernetesNodeSelector,
       image,
       dockerVolumes
     });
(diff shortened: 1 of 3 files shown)


dacbd:
@ludelafo can you provide an simple example command of at least 2 cases/values a user list:
```
cml runner launch \
    ... \
    --cloud-kubernetes-nod-selector xxx
```

dacbd (changes requested):

bin/cml/runner/launch.js
```diff
@@ -588,6 +590,20 @@ exports.options = kebabcaseKeys({
     description: 'Specifies the subnet to use within AWS',
     alias: 'cloud-aws-subnet-id'
   },
+  cloudKubernetesNodeSelector: {
+    type: 'string',
```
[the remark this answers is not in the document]
ludelafo: An array makes more sense indeed. I'll update the PR to reflect this.

bin/cml/runner/launch.js
```diff
@@ -588,6 +591,17 @@ exports.options = kebabcaseKeys({
     description: 'Specifies the subnet to use within AWS',
     alias: 'cloud-aws-subnet-id'
   },
+  cloudKubernetesNodeSelector: {
+    type: 'array',
+    string: true,
+    default: ['accelerator=infer'],
```
ludelafo, line 597: > Can we keep the default value logic on the provided side?
> 
> I would prefer it match the other "provider" specific options and have a default value of `''` see `cloudAwsSubnet` as an example.

@dacbd, I tried to set the default value logic on the Terraform Provider Iterative side with no success.

I tried the following approaches:

```go
// https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#default
"kubernetes_node_selector": &schema.Schema{
	Type:     schema.TypeMap,
	ForceNew: true,
	Optional: true,
	Elem: &schema.Schema{
		Type: schema.TypeString,
	},
	Default: map[string]string{
		"accelerator": "infer",
	},
},
```

and

```go
// https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#defaultfunc
"kubernetes_node_selector": &schema.Schema{
	Type:     schema.TypeMap,
	ForceNew: true,
	Optional: true,
	Elem: &schema.Schema{
		Type: schema.TypeString,
	},
	DefaultFunc: func() (any, error) {
		return map[string]string{
			"accelerator": "infer",
		}, nil
	},
},
```

but every time I wasn't specifying the `--cloud-kubernetes-node-selector` argument to CML, the default value were just ignored on the provider side.

Have I missed something regarding this aspect?

For the moment, I left the default value on the CML side to be more explicit to the end user that the inference is made on the `accelerator` key.

dacbd: Sure, I mostly don't want to have yargs populate a default value. With how you had it, it would populate into the generated terraform even when it's not used, which for others, down the line, may cause confusion. (for example, why is there a k8s node selector when I'm using AWS).

dacbd: You can utilize this line to see what I mean: https://github.com/iterative/cml/blob/1be24edaa817de320a657ec3ad1182e145aecef7/src/terraform.js#L123

dacbd: with this line: i believe we are safe to remove the default value from the cml side: https://github.com/iterative/terraform-provider-iterative/pull/747/files#diff-08560ef5459da3ede56dd9700b81f84384dbff1de69918854a2b8e2da0155155R72

dacbd (approved):

dacbd:
@ludelafo, just a quick update for you, there seems to be an issue with our tests with one of our dependencies unrelated to any of the open PRs, and I want to make sure we have it resolved before building a new release. Unfortunately, it's not the highest priority for us ATM. @0x2b3bfa0 or I will try to resolve it soon, and we should be able to get this merged and a new release made immediately after.

Thanks for your patience.

ludelafo:
> @ludelafo, just a quick update for you, there seems to be an issue with our tests with one of our dependencies unrelated to any of the open PRs, and I want to make sure we have it resolved before building a new release. Unfortunately, it's not the highest priority for us ATM. @0x2b3bfa0 or I will try to resolve it soon, and we should be able to get this merged and a new release made immediately after.
> 
> Thanks for your patience.

No worries! Thanks for the update.

If I can be of any help, please let me know!

0x2b3bfa0:
Sorry for the delay, @ludelafo; tests were broken because of multiple reasons (#1382 & #1383) unrelated to this pull request. 🙈

Issue #1369, opened 2023-03-21
`cml publish` ignores paths with spaces


lainisourgod:
I use `cml comment create --publish-native` to publish report with images. All images are uploaded correctly except images with with spaces in paths: `result/Some category/plot.png`. 

CML transformed my markdown`![]( result/Some category/pr_curve.png)` to `!\[]\( result/Some category/pr_curve.png)` (added some backslashes for some reason, didn't upload file).

When I transformed my path to contain `%20` instead of spaces, `![]( result/Some%20category/pr_curve.png)`, CML just put this in comment as is.

kshitiz305:
HI Team I would like to contribute to this issue.

kshitiz305:
Can we type the path in this manner that will allow it to work       

`![]( "result/Some category/pr_curve.png")`

PR #1373, opened 2023-04-09
handle paths with spaces

Autodidact24:
Resolving #1369 

We can use the `encodeURIComponent` function to encode any spaces in the `node.url` property before passing it to the `path.resolve` function.

This should ensure that image URLs with spaces are resolved correctly and passed to the publish function.


0x2b3bfa0:
Thanks, @Autodidact24! Can you please add a test case for the changes introduced on this pull request? Sorry for the late reply.

0x2b3bfa0:
By the way, why does `path.resolve` need arguments to be URL-encoded? 🤔

Issue #1403, opened 2023-07-19
New Feature


nipierre:
## Summary / Background

The GitLab driver is using a deprecated environment variable to fetch branch name (namely `CI_BUILD_REF_NAME`).

## Scope

Users using CML on GitLab 16.x.

## Assumptions

Should be replaced by either `CI_COMMIT_REF_NAME` or `CI_COMMIT_BRANCH`.

## Open Questions

-

## Blockers / Dependencies

-

## General Approach

In GitLab CI:
`cml pr .`

Ends up with:
```
{"level":"error","message":"git,checkout,-B,,c8ea18db3c47a566b6939be8b8c2831dbc5665df\n\t\n\tfatal: '' is not a valid branch name\n","stack":"Error: git,checkout,-B,,c8ea18db3c47a566b6939be8b8c2831dbc5665df\n\t\n\tfatal: '' is not a valid branch name\n\n    at /usr/lib/node_modules/@dvcorg/cml/src/utils.js:38:18\n    at ChildProcess.exithandler (node:child_process:410:5)\n    at ChildProcess.emit (node:events:513:28)\n    at maybeClose (node:internal/child_process:1100:16)\n    at Process.ChildProcess._handle.onexit (node:internal/child_process:304:5)"}
```

## Timelines

I can push a fix by the end of the week.

dacbd:
@nipierre good catch! @ mention me when you create your PR and I'll prioritize getting it merged and released

PR #1404, opened 2023-07-20
fix: replace deprecated gitlab ci var

nipierre:
GitLab 16.x deprecated several variables including CI_BUILD_REF_NAME. In order to be exhaustive, two variables are replacing it: CI_COMMIT_BRANCH (prefered) and CI_COMMIT_REF_NAME.

Issue: #1403

I put the two variables in order to cover most of the cases one can think of in terms of pipeline trigger.

@dacbd

### DIFF
--- a/src/drivers/gitlab.js
+++ b/src/drivers/gitlab.js
@@ -530,7 +530,10 @@ class Gitlab {
   }
 
   get branch() {
-    return process.env.CI_BUILD_REF_NAME;
+    if ('CI_COMMIT_BRANCH' in process.env) {
+      return process.env.CI_COMMIT_BRANCH
+    }
+    return process.env.CI_COMMIT_REF_NAME;
   }
 
   get userEmail() {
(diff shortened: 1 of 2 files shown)


src/drivers/gitlab.js
```diff
@@ -530,7 +530,10 @@ class Gitlab {
   }
 
   get branch() {
-    return process.env.CI_BUILD_REF_NAME;
+    if ('CI_COMMIT_BRANCH' in process.env) {
+      return process.env.CI_COMMIT_BRANCH
```
dacbd, line 534:
```suggestion
      return process.env.CI_COMMIT_BRANCH;
```

nipierre:
@dacbd I fixed the credentials build missing argument.

dacbd:
@nipierre Thanks, I'm addressing a container issue that is blocking some of our automation, my plan is to try and have this merged and released by the end of the week for you.

Thanks again for this patch!

dacbd (approved):

PR #1429, opened 2023-10-02
pin last `vega-lite` version to support node16

dacbd:


shcheklein:
Thanks @dacbd !

- Is it a big undertaking to upgrade those images?
- Could you please remind - what is the difference on using them vs `setup-cml`. What is the main use case? Do we want to keep maintaining them?

dacbd:
This patch is a relatively simple undertaking.

As far as maintaining both, I would consider the docker images to be in maintenance mode. The thing that bothers me the most is that the cml release workflow is very fragile in its current state, so creating as few releases as possible is my preference.
The last few releases have always had some issues.


`setup-cml` is much easier to maintain, and is what I always recommend. The couple of times users have wanted a container image I've recommended they take a look at our Dockerfile in the repo as an example to create their own image with everything they need for their use-case.

IMO if you are using docker images for your project pipelines you shouldn't use any `setup-*` style actions. Those tools should be built into your image with pinned versions so you have the exact stable environment you want/need.

shcheklein:
Is setup-cml available only for GH though? Pretty much in general people have to use cml images, no? btw, do we have the `dvc3` version of it (probably need to update the website and docs)

dacbd:
> Is setup-cml available only for GH though? Pretty much in general people have to use cml images, no? btw, do we have the `dvc3` version of it (probably need to update the website and docs)

Yes BB pipelines and gitlab-ci don't have anything similar (to setup-cml) and are essentially based on containers.

People don't have to use the CML images still they can install cml into with own image, or use @0x2b3bfa0's `curl https://cml.sh/ | sh`. 
Personally, I've always viewed the cml docker images as "getting started" tools that have a nice bundle of tools to help get you started.

Yes we build`dvc3` images also, if the docs do reflect that we should probably update them.

shcheklein:
okay, fine then ... let's pin the images and probably bump the website to use dvc3. No need to spend too much time on it for now. I'm a bit worried that Node 16 is out of support now. So, probably we should put a note somewhere on the website as well that we don't recommend using those images - they are just an example.

shcheklein (approved):

shcheklein:
@0x2b3bfa0 @dacbd do we know why it keeps creating comments here? :)

dacbd:
@shcheklein I think its one of the tests, this was the last PR merged

Issue #1478, opened 2024-09-27
CML comment create/update image links not working anymore in Gitlab


jancrichter:
Hi everyone,

We are using Gitlab 17.4.1 and CML 0.20.4. We have recently started noticing that images added to MR comments via `cml comment create` or `cml comment update`, while uploaded correctly, cannot be accessed under the link CML puts into the MR comment.

CML puts in absolute URLs like

```
![](https://${myinstance}.gitlab.com/${mygroup}/${mysubgroup}/${myproject}/uploads/6aaacfcf584e185f78b6df984148ad6a/plot.png?cml=png&cache-bypass=b4c2a1c9-0d6b-41b4-975b-168e1d150e12 "plot_title")
```
For new file uploads in Gitlab 17 these URLs do not work anymore (they continue working for uploads that already exist).

What works instead is this:

```
![](https://${myinstance}.gitlab.com/-/projects/${myprojectsid}/uploads/6aaacfcf584e185f78b6df984148ad6a/plot.png?cml=png&cache-bypass=b4c2a1c9-0d6b-41b4-975b-168e1d150e12 "plot_title")
```
I was not able to find a change in Gitlab 17 that would explain this behaviour yet, so it might be configuration on our end. I talked to the admins but they aren't aware of anything either.

I would like to propose switching to a more robust version of the links though.
When users manually upload files to a comment, Gitlab simply puts in a relative path like this:

```
![](/uploads/6aaacfcf584e185f78b6df984148ad6a/plot.png?cml=png&cache-bypass=b4c2a1c9-0d6b-41b4-975b-168e1d150e12 "plot_title")
```

Is there any reason not to use this approach?

shcheklein:
> Is there any reason not to use this approach?

Seems reasonable. @skshetry do you have any information on this?

PR #1482, opened 2024-10-24
Use base URL of GitLab for the `/uploads` URLs

0x2b3bfa0:
Could close https://github.com/iterative/cml/issues/1478; not sure about compatibility with old GitLab versions, though.

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -59,6 +59,11 @@ jobs:
           | jq -r .id \
           | xargs -0 printf "hash=%s" >> $GITHUB_OUTPUT
       - run: npm ci
+      - uses: mxschmitt/action-tmate@v3
+        with:
+          limit-access-to-actor: true
+        env:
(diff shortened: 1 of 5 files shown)


0x2b3bfa0:
@jancrichter, if you can test it with older GitLab versions and it works as well, we can release this fix.

> ![NOTE]
> It's a bit hacky, as it doesn't output full URLs unlike the assets.cml.dev uploads, but... 🤷🏼

jancrichter:
Hi, thanks for looking into it. Although, I am confused due to the PR being closed and the branch deleted, is it still an option to release this?

Myself, I cannot test on older Gitlab versions unfortunately, but the shortened link is what Gitlab has been using at least since Gitlab 15 from looking at the API.

shcheklein (approved):

0x2b3bfa0:
@jancrichter, yes, I just closed it because I wasn't willing to merge this change without your confirmation, but if it runs on GitLab 15+, it feels like a good enhancement to me.

(Now it's merged and released, for the record)

